PR #6, opened 2018-03-10
Master cpu measurements

amete:
Adding CPU time measurements to the outputs.

### DIFF
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -8,15 +8,17 @@
 #include "rapidjson/stringbuffer.h"
 #include "rapidjson/filewritestream.h"
 #include <math.h>
+#include <unistd.h>
 
 using namespace rapidjson;
 
-int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valuesIO[4], bool verbose){
+int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valuesIO[4], unsigned long long valuesCPU[4], bool verbose){
 
   //Get child process IDs
       std::vector<pid_t> cpids;
       char smaps_buffer[64];
       char io_buffer[64];
+      char stat_buffer[64];
       snprintf(smaps_buffer,64,"pstree -A -p %ld | tr \\- \\\\n",(long)mother_pid);
       FILE* pipe = popen(smaps_buffer, "r");
       if (pipe==0) {
@@ -52,8 +54,14 @@ int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valu
       unsigned long long trbyte(0);
       unsigned long long twbyte(0);
 
+      unsigned long utime(0);
+      unsigned long stime(0);
+      unsigned long cutime(0);
+      unsigned long cstime(0);
+
       std::vector<std::string> openFails;
 
+      char sbuffer[2048], *tsbuffer;
       for(std::vector<pid_t>::const_iterator it=cpids.begin(); it!=cpids.end(); ++it) {
         snprintf(smaps_buffer,64,"/proc/%lu/smaps",(unsigned long)*it);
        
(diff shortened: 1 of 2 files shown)


package/src/prmon.cpp
```diff
@@ -8,15 +8,17 @@
 #include "rapidjson/stringbuffer.h"
 #include "rapidjson/filewritestream.h"
 #include <math.h>
+#include <unistd.h>
 
 using namespace rapidjson;
 
-int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valuesIO[4], bool verbose){
+int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valuesIO[4], unsigned long long valuesCPU[4], bool verbose){
```
graeme-a-stewart: Small point (and not your error), but `mother_pid` and `verbose` could be `const`.

package/src/prmon.cpp
```diff
@@ -84,9 +92,29 @@ int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valu
...
+        }
+        else {
+          long clock_ticks = sysconf (_SC_CLK_TCK);
+          while(fgets(sbuffer,2048,file3)) {
+            tsbuffer = strchr (sbuffer, ')');
+            if(sscanf(tsbuffer + 2 , "%*c %*d %*d %*d %*d %*d %*u %*u %*u %*u %*u %80llu %80llu %80llu %80llu", &utime, &stime, &cutime, &cstime)) {
+              valuesCPU[0] = utime/clock_ticks;
+              valuesCPU[1] = stime/clock_ticks;
+              valuesCPU[2] = cutime/clock_ticks;
+              valuesCPU[3] = cstime/clock_ticks;
```
graeme-a-stewart: I would not divide by `clock_ticks` here. We should keep resolution until we need to drop it (plus as your are doing integer division, you round down to the nearest second unnecessarily).

package/src/prmon.cpp
```diff
@@ -84,9 +92,29 @@ int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valu
...
 
+        snprintf(stat_buffer,64,"/proc/%llu/stat",(unsigned long long)*it);
+
+        FILE *file3 = fopen(stat_buffer,"r");
+
+        if(file3==0) {
+          openFails.push_back(std::string(stat_buffer));
+        }
+        else {
+          long clock_ticks = sysconf (_SC_CLK_TCK);
```
graeme-a-stewart: Don't need this here - see L110 comment.

package/src/prmon.cpp
```diff
@@ -158,7 +189,11 @@ int MemoryMonitor(pid_t mpid, char* filename, char* jsonSummary, unsigned int in
 	       << valuesIO[0]   << "\t" 
 	       << valuesIO[1]   << "\t" 
 	       << valuesIO[2]   << "\t" 
-	       << valuesIO[3]   << std::endl;  
+	       << valuesIO[3]   << "\t"
```
graeme-a-stewart: Divide by `clock_ticks` here instead (with conversion to `float` or `double`).

package/src/prmon.cpp
```diff
@@ -202,6 +240,13 @@ int MemoryMonitor(pid_t mpid, char* filename, char* jsonSummary, unsigned int in
 	  }
           tmp = 0;
 
+          // Total CPU measurements
+          for(Value::MemberIterator it = v1.MemberBegin()+8; it != v1.MemberEnd(); ++it) {
+            it->value.SetUint64(maxValuesCPU[tmp]);
```
graeme-a-stewart: Divide by `clock_ticks` here, and output as float or double.

package/src/prmon.cpp
```diff
@@ -52,8 +54,14 @@ int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valu
       unsigned long long trbyte(0);
       unsigned long long twbyte(0);
 
+      unsigned long utime(0);
```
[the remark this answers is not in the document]
amete: Done.

package/src/prmon.cpp
```diff
@@ -84,9 +92,29 @@ int ReadProcs(pid_t mother_pid, unsigned long values[4], unsigned long long valu
...
+        FILE *file3 = fopen(stat_buffer,"r");
+
+        if(file3==0) {
+          openFails.push_back(std::string(stat_buffer));
+        }
+        else {
+          long clock_ticks = sysconf (_SC_CLK_TCK);
+          while(fgets(sbuffer,2048,file3)) {
+            tsbuffer = strchr (sbuffer, ')');
+            if(sscanf(tsbuffer + 2 , "%*c %*d %*d %*d %*d %*d %*u %*u %*u %*u %*u %80llu %80llu %80llu %80llu", &utime, &stime, &cutime, &cstime)) {
```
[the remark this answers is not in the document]
amete: This one I cannot reproduce. Maybe because I'm picking gcc73 via asetup? But I think this should go away once I properly define things as `unsigned long long` above. Please let me know if otherwise.

package/src/prmon.cpp
amete: Done.

package/src/prmon.cpp
amete: BTW I think your comment was for the `valuesCPU` bit and not the `valuesIO`.

graeme-a-stewart: Yes! Sorry for any confusion.

package/src/prmon.cpp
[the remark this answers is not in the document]
graeme-a-stewart: On CentOS7 I didn't get this, but did happen on Ubuntu16.04 with the gcc-7 compiler (which is odd). However, if I added the flag `-Wall` on CentOS7 it was reported.

We should actually get CMake to add that flag in all cases.

PR #10, opened 2018-03-14
Use getopt parsing for options

graeme-a-stewart:
boost::program_options started to look way to heavy for a simple
stand alone utility like this, so although it's a bit more work to
use getopt, it makes the build a lot easier.

Remove the couple of places where a C++14 construct was
used, to drop the required C++ version to 11. This allows building
with compilers as old as gcc4.8 (e.g., native on CentOS7).

Fixed a bug in the way the the RapidJSON include file was passed.

Removed doxygen build elements from the CMakeLists files as
this project does not need them.

Update README.

### DIFF
--- a/package/tests/burner.cpp
+++ b/package/tests/burner.cpp
@@ -14,8 +14,6 @@
 
 #include "burner.h"
 
-const float default_runtime = 10.0;
-
 double burn(unsigned long iterations = 10000000lu) {
   // Perform a time wasting bit of maths
   // Use volatile to prevent the compiler from optimising away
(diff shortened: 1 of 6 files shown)


package/tests/burner.cpp
```diff
@@ -41,33 +39,49 @@ double burn_for(float ms_interval = 1.0) {
...
-      return 1;
+  if (do_help) {
+    std::cout <<
+        "burner is a simple cpu burner program that can run in multiple threads\n"
+        "and/or processes.\n\n"
+        "If both threads and procs are sepecified then each process runs\n"
+        "multiple threads (so the load is threads * procs).\n" << std::endl;
+    std::cout << "Options:\n"
+        << " [--threads, -t N]  Number of threads to run (default 1)\n"
+        << " [--procs, -p N]    Number of processes to run (default 1)\n"
```
amete: A minor point but the default values for threads and processes are set to the hardware concurrency value instead of 1. Maybe we should document as such.

package/tests/burner.cpp
```diff
@@ -41,33 +39,49 @@ double burn_for(float ms_interval = 1.0) {
...
+        << " [--time, -r T]     Run for T seconds (default " << default_runtime << "\n"
```
amete: I noticed that if the user doesn't provide a time value, it's actually set to zero and not the default runtime value of 10 secs as intended:

$ ./build/package/tests/burner 
Will run for 0s using 10 process(es) and 10 thread(s)

graeme-a-stewart:
Nicely spotted, thanks!

I improved a lot now the handling of default values and also implemented an early exit when an unknown option is encountered

amete:
Looks good, many thanks. Merging now.

PR #12, opened 2018-03-16
Removed extra tabs from the header row

protopopescu:
Removed extra tabs between Time and VMEM columns in the header row. I use d3.tsv to parse the data, and the extra tabs are upsetting the JS parser. See http://memmon.ppe.gla.ac.uk/index.php?display=prmon


amete:
This looks good to me, I'm merging now.

protopopescu:
Great! Thank you very much!

PR #14, opened 2018-03-16
Improve CPU time monitoring unittest

graeme-a-stewart:
Write CPU monitor test in python as it's far easier to
check the results are sane. Add 3 tests to the CMakefile
that test single process, multi-process and multi-thead
running.

There is a bit of a weakness in the lower bound on CPU time,
as on a very busy machine it's possible that the CPU consumed
could be low, but OTOH it is a useful test that we do get some
roughly sane results.

### DIFF
--- a/package/tests/testCPU.py
+++ b/package/tests/testCPU.py
@@ -15,10 +15,10 @@ def setupConfigurableTest(threads=1, procs=1, time=10, slack=0.75):
     class configurableProcessMonitor(unittest.TestCase):
         def test_runTestWithParams(self):
             burn_cmd = ['./burner', '--time', str(time)]
-            if procs != 1:
-                burn_cmd.extend(['--procs', str(procs)])
             if threads != 1:
                 burn_cmd.extend(['--threads', str(threads)])
+            if procs != 1:
+                burn_cmd.extend(['--procs', str(procs)])
             burn_p = subprocess.Popen(burn_cmd, shell = False)
     
             prmon_cmd = ['../prmon', '--pid', str(burn_p.pid)]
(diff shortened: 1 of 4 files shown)


package/tests/testCPU.py
```diff
@@ -0,0 +1,49 @@
...
+    
+            burn_rc = burn_p.wait()
+            prmon_rc = prmon_p.wait()
+    
+            self.assertEqual(prmon_rc, 0, "Non-zero return code from prmon")
+            prmonJSON = json.load(open("prmon.json"))
+            totCPU = prmonJSON["Max"]["totUTIME"] + prmonJSON["Max"]["totSTIME"]
+            self.assertLess(totCPU, time*threads*procs, "Too high value for CPU time")
+            self.assertGreater(totCPU, time*threads*procs*slack, "Too low value for CPU time")
+
```
amete: I think somewhere around here it would be nice to have a simple printout of the values `totCPU` and `time*threads*procs` w/ and w/o `*slack` such that the user will have a bit more information (might be useful especially if the test fails).

package/tests/testCPU.py
```diff
@@ -0,0 +1,49 @@
...
+            self.assertLess(totCPU, time*threads*procs, "Too high value for CPU time")
+            self.assertGreater(totCPU, time*threads*procs*slack, "Too low value for CPU time")
+    
+    return configurableProcessMonitor
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="Configurable test runner")
+    parser.add_argument('threads', type=int)
+    parser.add_argument('procs', type=int)
```
amete: Very very minor but I think burner has the convention of first procs then threads (in printing), which is reversed here. Maybe we can adopt the same order to avoid possible confusion:

$ python2.7 testCPU.py 2 3 
Will run for 10s using 3 process(es) and 2 thread(s)

amete: Do we also want the user to have control over `time`?

package/tests/testCPU.py
```diff
@@ -0,0 +1,49 @@
...
+    args = parser.parse_args()
+    # Stop unittest from being confused by the arguments
+    sys.argv=sys.argv[:1]
```
amete: If the user doesn't provide any arguments we get:

$ python2.7 testCPU.py     
usage: testCPU.py [-h] threads procs
testCPU.py: error: too few arguments

Can we instead bail simply to `testCPU.py -h`?

package/tests/testCPU.py
graeme-a-stewart: The failure message should be more informative now (of course, all output is hidden at the `make test` stage, but it's in `Testing/Temporary/LastTest` and if run "by hand" you see it)

package/tests/testCPU.py
graeme-a-stewart: I try to be consistent in processing threads then procs.

Now all the arguments can be set, with optional sensible defaults (it's also then clearer for the user what test has been configured).

graeme-a-stewart:
OK, should all be improved - thanks for the helpful comments.

I worry more about ensuring we work with python3 than caring about python 2.6, but it might be possible to be ok for 2.6 (which is the SLC6 native, right?) as long as it supports our imports from `__future__`.

graeme-a-stewart:
Ah, I tested on lxplus7, and things do work fine as the default python is 2.7.5.

On SLC6 both gcc and python are too old, so I don't care to support them "out of the box" (one can use the LCG builds to supply the requisite dependencies there).

amete:
Perfect, thanks a lot @graeme-a-stewart! I agree w/ the native SLC6. Both gcc and python are pretty old, and in any real life usage people will end up picking up more recent versions from LCG anyways. Merging now.

Issue #16, opened 2018-03-16
Add an IO burner program and tests


graeme-a-stewart:
This is to test prmon against a 'standard candle' on IO loads.

graeme-a-stewart:
Development in https://github.com/HSF/prmon/tree/io-burner branch

amete:
Fixed in #17, closing.

PR #17, opened 2018-03-17
Io burner and unit tests

graeme-a-stewart:
Implements an io-burner test program and unittests for the same. Addresses #16.

Should be good to go, but may be need rebased after #14 is accepted.


graeme-a-stewart:
Patch has been rebased, so this is now ready to accept.

amete:
This all looks good and works as expected @graeme-a-stewart. I'm merging, thanks for the rebase :)

PR #20, opened 2018-03-21
First implementation of wall-time calculation

amete:
This is the initial implementation of getting the wall-time for the monitored process. Tried to keep it consistent w/ everything else, we're reading all the relevant information from `/proc`.

### DIFF
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -154,9 +154,9 @@ int MemoryMonitor(const pid_t mpid, const std::string filename, const std::strin
      // Open iteration output file     
      std::ofstream file;  
      file.open(filename);
-     file << "Time\tVMEM\tPSS\tRSS\tSwap\trchar\twchar\trbytes\twbytes\tutime\tstime\tcutime\tcstime" << std::endl;
+     file << "Time\tVMEM\tPSS\tRSS\tSwap\trchar\twchar\trbytes\twbytes\tutime\tstime\tcutime\tcstime\twtime" << std::endl;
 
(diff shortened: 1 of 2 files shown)


amete:
Thanks @graeme-a-stewart. I added a simple test to check the wall-time measurement. 

I also noted the same thing and agree that 1 sec offset is OK for all intents and purposes. I can look into improving this in the future. 

I also noted that even if we specify `--interval 1` to `prmon` we end up w/ measurements that are 2 secs apart (might also be machine dependent). At some point, I'm also planning to take a look to see if we can speed things up a bit more to improve the resource usage on the `prmon` side.

graeme-a-stewart:
Hi @amete - now it looks good. However, I am seeing core dumps in my Ubuntu container. Let me try and figure out why before merging this.

amete:
Thanks a lot @graeme-a-stewart, let me know :)

graeme-a-stewart:
Yes, it's real - the current head of master works fine. So some regression I need to check...

package/src/prmon.cpp
```diff
@@ -173,10 +173,56 @@ int MemoryMonitor(const pid_t mpid, const std::string filename, const std::strin
...
+           }
+         }
+       }
+       fclose(stat_file);
+     }
+
+     // Get system uptime
+     snprintf(stat_buffer,64,"/proc/uptime");
+
+     fopen(stat_buffer,"r");
```
graeme-a-stewart: Bug! Should be

`stat_file = fopen(stat_buffer,"r");`

amete: Wow! I'm a bit surprised how it actually worked fine w/ this bug. Should be fixed now.

PR #22, opened 2018-03-22
Switch on stricter compiler checks

graeme-a-stewart:
Enable stricter checks in the compiler to make sure that we don't
have warnings - indeed treat them as errors.

Tidy up code in the places where warning conditions had crept in.

@amete - this would have caught that bug that almost came in on #20 
so it's worth doing.

I reworked the `io-burner` code a bit, but it's worth a second pair of 
eyes checking for any buffer overruns.


amete:
All looks good, merging :)

PR #28, opened 2018-03-30
Network stats added

### LINKED ISSUE #2: Gather network statistics from prmon
Not hugely clear how to get this per-process, but measuring "global" network activity in a controlled process space might be sufficient for this purpose (docker, cgroups?).

graeme-a-stewart:
This adds network statistics monitoring to resolve #2.

Stats are gathered from all `/sys/class/net/DEVICE/statistics`, measuring `rx_bytes`, `rx_packets`, `tx_bytes`, `tx_packets`.

To have a unittest a small python program was written that requests data from a python localhost http server, started by the test harness. As there's no way to know what other network traffic is happening only the *minimum* expected amount of traffic is tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ most useful features is to use smaps to correctly calculate the
 is a much better indication of the true memory consumption of
 a group of processes where children share many pages.
 
-prmon currently runs on linux machines as it requires access to the
+prmon currently runs on Linux machines as it requires access to the
 `/proc` interface to process statistics.
 
 ## Build and Deployment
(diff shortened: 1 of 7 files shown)


amete:
Hi @graeme-a-stewart,

These look good and I'm happy to merge them. Before doing so, just for my own education, we're essentially listening to the transfers through `eth0` `eth1` etc. but we don't really know if these are associated to the `proc` that we're interested in, is that so? If so, do you think we can extract some `proc` specific information from `/proc/[pid]/net/netstat`? 

Best,
Serhan

graeme-a-stewart:
Right - we don't know. So one option would be to allow the user to select which network interfaces to monitor. That could be added reasonably easily, I think, e.g.,

```sh
--dev eth0,lo
```

amete:
Thanks @graeme-a-stewart, sounds good. Please let me know if you'd like to do it in this MR (in which case I can wait to merge) or handle it in a separate one.

-s

graeme-a-stewart:
I'll add a little patch now, so expect another commit soon...

graeme-a-stewart:
P.S. I did not answer your question properly. `/proc/PID/net/netstat` is actually the same for all processes, so, no, there is no way on a vanilla kernel to get per-process network I/O stats.

Programs like [nethogs](https://www.tecmint.com/nethogs-monitor-per-process-network-bandwidth-usage-in-real-time/) can do it, but they need `root`.

In the end it was easier to specify multiple network devices with multiple `--netdev` or `-n` options. Splitting an `std::string` into multiple pieces is remarkably awkward in C++.

PR #29, opened 2018-04-03
Simple python script to visualize proc output

### LINKED ISSUE #25: Visualization script
I was thinking it might be good to add a script that produces basic 2-D plots that'll help user visualize resource usage (CPU, memory etc.) as a function of time out of `prmon` output. Can be easily done using commonly used libraries, like `matplotlib`.

amete:
Targets to resolve #25.


amete:
Thanks @graeme-a-stewart. I renamed the script as you suggested and made a few tweaks to `CMakeLists.txt`s such that it gets installed under `/bin`. However, I can change the latter behaviour. Please let me know, thanks.

Issue #27, opened 2018-03-30
Allow prmon to invoke the monitored program itself


graeme-a-stewart:
A very nice feature extension for `prmon` would be to be able to fork off the program that should be monitored, as alternative to passing the `PID` of an existing process.

Use would be like,

```sh
prmon --json-summary myout.json --interval 5 -- ./monitored_program --progarg1 1 --progarg2 2
```

using the traditional `--` to separate `prmon` arguments from the child's arguments.

amete:
I have a first shot at this in https://github.com/HSF/prmon/tree/master-self-invoke

I do a simple `fork`, get the `pid` of the `child` where we execute the monitored program, use that in the `parent` to listen. The usage is like:

`prmon --args "exec arg1 arg2 arg3"`

I works to the first order but not finalized yet. Since we loop in the `MemoryMonitor` w/o an appropriate `wait` in the `parent` for the `child` the latter becomes a zombie therefore we keep listening indefinitely. I'll take a look to find a nice way of getting around this but any comments are welcome :)

graeme-a-stewart:
Hi @amete - thanks, good start.

I would use the more standard way to invoke a child process with the "--" though. I think that can be easily pre-parsed off argv before we parse other arguments.

Regarding the child hanging around for ever, prmon will get a `SIGCHLD` when it has exited. We just trap that and read the exit status to kill the zombie. I am sure I have done this in the distant past...

amete:
Thanks @graeme-a-stewart.

I dropped the `--args` and switched to `--` approach as you suggested. BTW great tip w/ the `SIGCHLD`, thanks! After a bit of searching on the internet and experimenting :) I think I no have a version that is actually quitting upon termination of the child process.

Could you take a look, give it a shot and let me know, please? Many thanks.

graeme-a-stewart:
Hi @amete - Thanks. I made some more improvements and just pushed them to the branch. Commit messages have the details.

I think we just need to add a few tests and this is ready to go.

graeme-a-stewart:
I think modifying an existing test (burner?) would be just fine to test that we can invoke external programs properly.

amete:
I updated `package/tests/testCPU.py` w/ a flag `--invoke` that runs the CPU test via the new prmon functionality using `--` (also added a new test `testCPUinvoke`). I think this should serve the purpose but let me know.

graeme-a-stewart:
[...]

PR #30, opened 2018-04-13
Allow prmon to invoke the monitored program itself

amete:
Enhancements discussed at #27.


package/src/prmon.cpp
```diff
@@ -2,12 +2,15 @@
   Copyright (C) 2018, CERN
 */
 
+#include <algorithm>
```
graeme-a-stewart: I don't think we need this...

amete: Good catch, I was playing around and this a left over from that :) I took that out.

graeme-a-stewart: Ha, yes, it's easy on the diff, but we should probably run IWYU properly too

Issue #31, opened 2018-04-13
Refactor monitoring to "plugins"


graeme-a-stewart:
As was noted in #2 the structure of `prmon` makes it difficult to easily add or remove parts of the monitoring.

I think this could be refactored along the following lines:

- Each piece of the monitoring is a class instance (`cpumon`, `netmon`, `iomon`, ...)
- Each of these classes implements an monitoring interface that provides the standard components need by `prmon` (e.g., print text file headers, get statistics, print JSON entries)

Then `prmon` can keep a vector of the enabled instances and just loop over them to assemble the requisite monitoring in a flexible way.

graeme-a-stewart:
Fixed in #37, although at the moment no runtime switching of monitoring modules is done, it's much easier to support this now.

PR #33, opened 2018-04-13
Netmon refactor

graeme-a-stewart:
This is a refactoring of the network monitoring. I would not consider it a finished version (e.g., to satisfy #31) but it's enough of an improvement to merit merging now.

### DIFF
--- a/package/src/netmon.h
+++ b/package/src/netmon.h
@@ -47,8 +47,7 @@ class netmon {
     return filename;
   }
 
-  // Internal methods to read "raw" network stats
-  std::unordered_map<std::string, unsigned long long> read_raw_network_stats();
+  // Internal method to read "raw" network stats
   void read_raw_network_stats(
       std::unordered_map<std::string, unsigned long long>& values);
 
@@ -64,10 +63,7 @@ class netmon {
     return monitored_netdevs;
   }
 
-  // Return network stats in a variable
-  std::unordered_map<std::string, unsigned long long> read_network_stats();
-
-  // Return network stats by modifying a reference
+  // Return network stats
   void read_network_stats(
       std::unordered_map<std::string, unsigned long long>& values);
 };
--- a/package/src/netmon.h
+++ b/package/src/netmon.h
@@ -17,7 +17,7 @@
 #include <vector>
 
 class netmon {
-private:
+ private:
   // Which network interface paramters to measure
   const std::vector<std::string> interface_params;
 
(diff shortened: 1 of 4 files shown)


package/src/netmon.h
```diff
@@ -0,0 +1,75 @@
...
+
+  // Nested dictionary of network_if_streams[PARMETER][DEVICE][ISTREAM*]
+  std::unordered_map<
+      std::string,
+      std::unordered_map<std::string, std::unique_ptr<std::ifstream>>>
+      network_if_streams;
+
+  // Container for starting value stats, that are subtracted from measured
+  // values
+  std::unordered_map<std::string, unsigned long long> network_stats_start;
```
amete: `const` maybe? I don't think we're planning to change this throughout the program execution anyhow.

package/src/netmon.h
```diff
@@ -0,0 +1,75 @@
...
+
+  // Helper to map a device and parameter to a /sys filename
+  inline std::string const get_sys_filename(std::string device,
+                                            std::string param) {
+    std::string filename = "/sys/class/net/" + device + "/statistics/" + param;
+    return filename;
+  }
+
+  // Internal methods to read "raw" network stats
+  std::unordered_map<std::string, unsigned long long> read_raw_network_stats();
```
amete: Similar to the comment below. In this case we actually use this guy to store the initial values which are used as an offset. Shall we make it a `get` method that returns a `const` so that the result doesn't get changed mistakenly?

package/src/netmon.h
```diff
@@ -0,0 +1,75 @@
...
+  const std::vector<std::string> get_interface_paramter_names() {
+    return interface_params;
+  }
+
+  const std::vector<std::string> get_interface_names() {
+    return monitored_netdevs;
+  }
+
+  // Return network stats in a variable
+  std::unordered_map<std::string, unsigned long long> read_network_stats();
```
amete: Do we really need this overloaded definition of `read_raw_network_stats `? I don't think we use this anywhere. If we want to have this as means to `read`-only, shall it be a `get` method that returns a `const` instead? Maybe I'm missing something?

package/src/netmon.h
graeme-a-stewart: But it has to be set to the correct values, so it can't be `const` here. Also, it may be that for error handling (like interfaces failing) we need to update it.

package/src/netmon.h
graeme-a-stewart: But if you return `const` then it can't then be passed as a modifiable reference to be updated by `read_network_stats(&value)`, which is what's done later.

OTOH, I start to not like too much the overload either, especially as we have the public methods as well as the private "raw" versions. The client could just create the container itself and pass it in to the single updater method.

package/src/netmon.h
amete: OK, I see now that this was a dubious comment. I was thinking we could utilize the initialization list to do the assignment at run-time but I see we compute what we pass to this guy in the constructer.

package/src/netmon.h
amete: Definitely, `const`ing them will break passing things around by reference for updating. I thought we might mistakenly update something and trying to see if there is a way to safeguard against that. Also I always find avoiding overloaded methods ease following the code but maybe that's personal :)

PR #34, opened 2018-04-21
Improve cpu burner to have short lived children

graeme-a-stewart:
Add a new paramater that runs the children for a shorter time than
the parent process, to ensure that the time for already exited children
has been properly accounted for by prmon

Make multi-proc test use short child lifetime

This commit makes the CPU monitoring test *fail*, because the way
it is currently implemnted loses the CPU time information from
children who exit before the mother process (as in the previous
incarnation of the test all processes exited at the same time
this was not noticed).

(The CPU time does become accounted for in cutime and cstime,
but I am not sure why we want these to be two separate fields - it's the total that's important.)

### DIFF
--- a/package/tests/CMakeLists.txt
+++ b/package/tests/CMakeLists.txt
@@ -38,7 +38,7 @@ script_install(SCRIPT httpBlock.py DESTINATION cgi-bin)
 # CPU Tests
 add_test(NAME testCPUsingle COMMAND testCPU.py --threads 1 --procs 1) 
 add_test(NAME testCPUmultithread COMMAND testCPU.py --threads 2 --procs 1) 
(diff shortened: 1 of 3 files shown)


amete:
Hi @graeme-a-stewart,

I tested this and all looks good to me. Nonetheless, I see the same changes are also in the #37. So, I've approved this but didn't merge it in. Let me know how you'd like to proceed.

Best,
Serhan

graeme-a-stewart:
I think let me know how you get on with #37 - if you think we can conclude on that this week then there's no need to bother about this minor point here.

graeme-a-stewart:
Closing as this improvement was folded into #37.

PR #36, opened 2018-04-23
Update the CPU usage report

amete:
The original implementation conserved the `/proc/stat` accounting, reporting parent and child CPU usage profiles separately. This meant that `utime` and `stime` were not necessarily monotonically increasing when a child exits earlier that the parent (in which case the associated usage was registered under `cutime` and `cstime`). We simply combine these to report a single monotonically increase value now (actually one for user and one for system usage).

### DIFF
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -32,7 +32,7 @@
 using namespace rapidjson;
 
 int ReadProcs(const pid_t mother_pid, unsigned long values[4],
-              unsigned long long valuesIO[4], unsigned long long valuesCPU[4],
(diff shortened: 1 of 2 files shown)


graeme-a-stewart:
Hi @amete - yes, I think this produces something which is more like the end user will want to see when plotting the `utime` and `stime`.

Though I tested this and it works, I have a monster rewrite which I am just about to ask you to review. So let's hold off merging this until you get a chance to take a look at that.

graeme-a-stewart:
We can close this now as it was folded into #37.

PR #37, opened 2018-04-23
Generic monitor re-implementation

graeme-a-stewart:
OK, this is a very big PR, so please take your time and have a proper look. I would look firstly at the design, before worrying about any of the details of the code.

The prosaic explanation for this is two very long bus journeys (Thursday and Sunday), with a lot of time to code, but no network access. So that I was able to go well beyond my prototype ideas for #31 and actually reimplement all of the monitoring components in the new generic way. Still it's a lot to swallow in one go and I'm sorry about that.

The core of the new design is a generic monitoring interface, `Imonitor`, which describes the interfaces needed for any monitoring component. 

- The design pattern is that the constructor is RAII, so just initialising a component gives one a "working" instance. 
- There is a polling method, `update_stats`, that takes a vector of PIDs, containing the child and parent process IDs (this is because many components need to loop over every active process). The component obviously updates its state when called.
- There are result `const` getter methods, `get_*`, which get the outputs of the monitor for each of the output *channels* that `prmon` has: viz., `get_text_stats` for the timestamped text file, `get_json_total_stats` for the JSON file `Max` dictionary and `get_json_average_stats` for the corresponding `Avg` dictionary. The latter takes a parameter, which is the number of *clock ticks* since the mother process started - this is needed for some of the monitors (the walltime monitor has a special method `wallmon::get_wallclock_clock_t` that supplies this value).

The output of all the getters is a `map` between a text key and an unsigned long long value. Each monitor can produce whatever key/value pairs it likes - they can be different for each channel (e.g., some monitors produce nothing for the JSON average channel). The keys can also be different, although in practice I did not do so in any component (in the end the final JSON structure of `value["Max"]["Param"]` does not need to to be `value["Max"]["MaxParam"]` IMO, although you might disagree - but the design can support both).

Note that there is a design *limitation* that all output values have to be of the same type, so that they can be handled uniformly. I could not think of any *easy* way to get a mixture of int types and float types and in the end I decided that as our use case is long lived monitoring, ensuring no truncation/rounding for very high values was more important than losing some precision for very low values.

In `prmon` all that is done now is to instantiate all of the monitors (using a vector of base class pointers), then each step of the monitoring is just a loop over each of the monitoring components.

I believe that this is a hugely better design as each monitoring component is self contained, making it much easier to see both what it's doing and how. It also allows the possibility of adding new monitors easily and of having runtime switches for enabling/disabling specific monitoring components.

The core `prmon.cpp` is now a lot easier to follow.

As a P.S. I am still not completely happy about how RapidJSON is used, it still seems clunky, making a string for the initial JSON structure and then dividing the document into `Max` and `Avg`. OTOH, there's enough in this PR already and I would really like your thoughts on it before rewriting absolutely **everything**.

### DIFF
--- a/package/tests/testCPU.py
+++ b/package/tests/testCPU.py
@@ -10,7 +10,7 @@ import subprocess
 import sys
 import unittest
 
-def setupConfigurableTest(threads=1, procs=1, time=10, slack=0.75, invoke=False):
+def setupConfigurableTest(threads=1, procs=1, child_fraction=1.0, time=10.0, slack=0.75, invoke=False):
     '''Wrap the class definition in a function to allow arguments to be passed'''
     class configurableProcessMonitor(unittest.TestCase):
         def test_runTestWithParams(self):
@@ -19,6 +19,8 @@ def setupConfigurableTest(threads=1, procs=1, time=10, slack=0.75, invoke=False)
                 burn_cmd.extend(['--threads', str(threads)])
             if procs != 1:
                 burn_cmd.extend(['--procs', str(procs)])
+            if child_fraction != 1.0:
+                burn_cmd.extend(['--child-fraction', str(child_fraction)])
 
             if invoke:
                 prmon_cmd = ['../prmon', '--']
@@ -39,10 +41,11 @@ def setupConfigurableTest(threads=1, procs=1, time=10, slack=0.75, invoke=False)
             prmonJSON = json.load(open("prmon.json"))
             # CPU time tests
             totCPU = prmonJSON["Max"]["utime"] + prmonJSON["Max"]["stime"]
-            self.assertLess(totCPU, time*threads*procs, "Too high value for CPU time "
-                            "(expected maximum of {0}, got {1})".format(time*threads*procs, totCPU))
-            self.assertGreater(totCPU, time*threads*procs*slack, "Too low value for CPU time "
-                               "(expected minimum of {0}, got {1}".format(time*threads*procs*slack, totCPU))
+            expectCPU = (1.0 + (procs-1)*child_fraction) * time * threads
+            self.assertLess(totCPU, expectCPU, "Too high value for CPU time "
+                            "(expected maximum of {0}, got {1})".format(expectCPU, totCPU))
+            self.assertGreater(totCPU, expectCPU*slack, "Too low value for CPU time "
+                               "(expected minimum of {0}, got {1}".format(expectCPU*slack, totCPU))
             # Wall time tests
             totWALL = prmonJSON["Max"]["wtime"]
             self.assertLessEqual(totWALL, time, "Too high value for wall time "
@@ -57,6 +60,7 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser(description="Configurable test runner")
     parser.add_argument('--threads', type=int, default=1)
     parser.add_argument('--procs', type=int, default=1)
+    parser.add_argument('--child-fraction', type=float, default=1.0)
     parser.add_argument('--time', type=float, default=10)
     parser.add_argument('--slack', type=float, default=0.75)
     parser.add_argument('--invoke', dest='invoke', action='store_true', default=False)
@@ -65,6 +69,6 @@ if __name__ == '__main__':
     # Stop unittest from being confused by the arguments
     sys.argv=sys.argv[:1]
     
-    cpm = setupConfigurableTest(args.threads,args.procs,args.time,args.slack,args.invoke)
+    cpm = setupConfigurableTest(args.threads,args.procs,args.child_fraction,args.time,args.slack,args.invoke)
     
     unittest.main()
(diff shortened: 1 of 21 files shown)


amete:
Thanks a lot @graeme-a-stewart. As a first comment I really like the structure. Having the base class `Imonitor` and separate classes for each metric that inherit from this makes things uniform and scalable. I'll checkout, play around and come back w/ more specific comments/questions regarding the implementation.

graeme-a-stewart:
Hi

No, I hadn't really seen anything like that - but I will also try to reproduce it. My main workhorse is a C7 docker container. Which environment are you working in?

We should also review failure modes, e.g., there are race conditions like the process exists right as prmon tries to read from `/proc/PID/stat` that would probably give a bad read (and should probably be just ignored).

Thanks

Graeme

graeme-a-stewart:
Yeah, and what the heck is signal 110!

amete:
I tried this on my desktop w/ `SLC6` plus both a) `cmake 3.11.0`,`gcc 4.9.3`, and `python 2.6.6` (which is archaic) and b) `cmake 3.11.0`, `gcc 6.2.0`, and `python 2.7.13` (which comes w/ 21.0 Athena nightly actually).

I thought 15 sec. would be enough to hook but maybe not?

amete:
I tried this on my desktop w/ `SLC6` plus both a) `cmake 3.11.0`,`gcc 4.9.3`, and `python 2.6.6` (which is archaic) and b) `cmake 3.11.0`, `gcc 6.2.0`, and `python 2.7.13` (which comes w/ 21.0 Athena nightly actually).

I thought 15 sec. would be enough to hook but maybe not?

PS : I mistakenly hit `close and comment` instead of `comment` - sorry about that :)

graeme-a-stewart:
Hi @amete 

I believe I figured it out - `waitpid()` will return immediately if the parent has no children that can be waited for. In that case the status doesn't get set, so it is processed uninitalised leading to very weird phantom exit codes. Now I really check that a child status was actually read properly. I also loop over `waitpid()` until there are no more children's statuses to be read, as it's possible to have a single `SIGCHLD` issued, but multiple children to be ready.

I found quite a few other rare bugs (for immediate exits of monitored processes and ifstream excptions), which are all fixed in the latest commits. This, this fixes #39 as well.

Please check again and let me know if you see more problems,

Thanks

Graeme

package/tests/testCPU.py
```diff
@@ -38,13 +40,14 @@ def test_runTestWithParams(self):
             self.assertEqual(prmon_rc, 0, "Non-zero return code from prmon")
             prmonJSON = json.load(open("prmon.json"))
```
amete: I just noticed this w/ a warning on my Docker/CentOS7 test. I believe we don't close this json file:

```
./testCPU.py:41: ResourceWarning: unclosed file <_io.TextIOWrapper name='prmon.json' mode='r' encoding='ANSI_X3.4-1968'>
  prmonJSON = json.load(open("prmon.json"))
```

Small point but would be good to close it properly (I think applies to other scripts as well).

package/tests/testNET.py
```diff
@@ -44,10 +44,10 @@ def test_runTestWithParams(self):
...
-            self.assertGreater(prmonJSON["Max"]["tot_rx_bytes"], expectedBytes, "Too low value for rx bytes "
-                               "(expected minimum of {0}, got {1})".format(expectedBytes, prmonJSON["Max"]["tot_rx_bytes"]))
-            self.assertGreater(prmonJSON["Max"]["tot_tx_bytes"], expectedBytes, "Too low value for tx bytes "
-                               "(expected minimum of {0}, got {1})".format(expectedBytes, prmonJSON["Max"]["tot_tx_bytes"]))
+            self.assertGreater(prmonJSON["Max"]["rx_bytes"], expectedBytes, "Too low value for rx bytes "
+                               "(expected minimum of {0}, got {1})".format(expectedBytes, prmonJSON["Max"]["rx_bytes"]))
+            self.assertGreater(prmonJSON["Max"]["tx_bytes"], expectedBytes, "Too low value for tx bytes "
+                               "(expected minimum of {0}, got {1})".format(expectedBytes, prmonJSON["Max"]["tx_bytes"]))
     
     return configurableProcessMonitor
```
amete: Here we use `urllib2` which makes this script python2 specific, if I'm not mistaken. In python3.6 it gives me:

```
  File "./netBurner.py", line 11, in <module>
    import urllib2
ModuleNotFoundError: No module named 'urllib2'
```

Also `CGIHTTPServer` is apparently moved into `http` module in python3. I just wanted to point out for future reference.

amete:
Thanks @graeme-a-stewart. I went over this as carefully as I could and after the last fixes all looks good to me. I put in two minor comments but these are not showstoppers. Therefore, I'm happy to merge.

package/tests/testCPU.py
graeme-a-stewart: I wonder why I never saw that myself? Usually I just let python files close themselves as they go out of scope, but this might be better as:

```python
with open("prmon.json") as jsonfh:
    prmonJSON = json.load(jsonfh)
    ...
```

We should create an issue to track this.

package/tests/testNET.py
graeme-a-stewart: Yeah, I really struggled to find a way to make this py2/3 neutral, but in the end I think I gave up and went for py2. The `six` module might save us:

https://pythonhosted.org/six/

Looking to the future, I would try to target python3, with python2 fallback.

graeme-a-stewart:
Excellent - thanks a lot. I am cleaning up some of the issues and other items we fixed here.

Issue #32, opened 2018-04-13
Improve child process parsing


graeme-a-stewart:
The way that child processes are found right now is pretty suboptimal, involving C-style text parsing of mangled `pstree` output. It's opaque and fragile.

A better way would be to interrogate `/proc` directly, from the parent process by finding children from 

`/proc/PID/task/PID/children`

and recursing.

Getting the vector of child PIDs should be a separate stage of the monitoring loop, with the PID vector then being passed to each of the monitoring plugins in #31.

graeme-a-stewart:
This is fixed in https://github.com/HSF/prmon/tree/new-child-pid-monitoring, but I will hold off on the PR until #37 is accepted as the changes are based on those.

PR #42, opened 2018-05-08
New child pid monitoring

### LINKED ISSUE #40: pstree can core dump
Strange, but true... at least I occasionally see this on Ubuntu 16 when extreme testing:

```
root@6af3cf4ea516:/tmp/prmon/package/tests# ../prmon -- ./burner -p 100 -c 0.1 -t 10
Will run for 10s using 100 process(es) and 10 thread(s)
Children will run for 1s
Segmentation fault (core dumped)
root@6af3cf4ea516:/tmp/prmon/package/tests# file core
core: ELF 64-bit LSB core file x86-64, version 1 (SYSV), SVR4-style, from 'pstree -A -p 1773'
```

It's harmless to `prmon` but would be weird for the user to see.

This will disappear when the https://github.com/HSF/prmon/tree/new-child-pid-monitoring branch is merged.

graeme-a-stewart:
Change the way we get child process ids, by looping over the information from /proc directly. This is more reliable than invoking `pstree` and doing painful string parsing in C.

Fixes #40 and #32.

### DIFF
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -2,9 +2,7 @@
   Copyright (C) 2018, CERN
 */
 
-#include <dirent.h>
 #include <getopt.h>
-#include <math.h>
 #include <signal.h>
 #include <stdlib.h>
 #include <sys/types.h>
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -13,6 +13,7 @@
 
 #include <condition_variable>
 #include <cstddef>
+#include <deque>
 #include <fstream>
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
Ping - this is ready for merge.

graeme-a-stewart:
Hi @amete - thanks for checking.

For me that test passes on C7 (local container), Ubuntu16 and on lxplus7 (run multiple times). I guess you are on lxplus6 in your case?

That debug output from `io-burner.cpp:204` should be removed.

PR #54, opened 2018-05-28
Reimplement old PID parser for older kernels

### LINKED ISSUE #47: prmon VMEM and PSS sizes seem to be wrong
Testing by @sciaba showed that there's a problem with VMEM and PSS. Evidently there's a bug in the re-implementation of the smaps parser in `memmon.cpp`.

Changes to `memmon` should be validated against the built-in `MemoryMonitor` run by ATLAS jobs.

graeme-a-stewart:
Switching to getting PIDs from /proc parsing in #42 broke the child PID tree finding on older kernels (particularly this affects SLC6) and makes the whole monitoring wrong.

This arises as old kernels do not have /proc exposure of their child PIDs. On these systems it's necessary to fall back to the previously used "pstree" parsing (this is less efficient and more fragile, but it does work well enough; https://lwn.net/Articles/475688/). 

To avoid testing on every iteration, a single test for the kernel's support of the new interface is done once, then a bool stored that switches between the implementations for old/new kernels in the monitoring loop.

There was enough different PID parsing code now to justify its own source file.

This fixes #47 and has been tested on SLC6 (as well as the usual C7 and U16 platforms).

### DIFF
--- a/package/CMakeLists.txt
+++ b/package/CMakeLists.txt
@@ -2,6 +2,7 @@ find_package(RapidJSON REQUIRED)
 find_package(Threads REQUIRED)
 
 add_executable(prmon src/prmon.cpp
+					 src/pidutils.cpp
 					 src/netmon.cpp 
 					 src/iomon.cpp 
 					 src/cpumon.cpp 
--- a/package/src/pidutils.cpp
+++ b/package/src/pidutils.cpp
@@ -1,0 +1,88 @@
+// Copyright (C) CERN, 2018
+
+// A few utilities handling PID operations
+
+#include <sys/stat.h>
+#include <stdlib.h>
+
+#include <cstddef>
+#include <deque>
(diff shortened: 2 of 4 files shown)


amete:
Hi @graeme-a-stewart,

Thanks a lot. I gave this a go on SLC6 and things look very good. I ran a q431 test on four cores over 100 events:

`MemoryMonitor`:

prmon_time_vs_vmem_pss_rss_memmon

`PrMon`:

prmon_time_vs_vmem_pss_rss_prmon

Results are ~identical. There seems to be a small spike in `MemoryMonitor` output that doesn't appear in the `PrMon` one, but I think that might be a brief spike that happens during HIST merging at the end that might have happened between the monitoring interval (30 secs) by chance (they are sampling a few seconds out of sync). 

Best,
Serhan

PR #61, opened 2018-05-30
Remove clock_ticks const statics

### LINKED ISSUE #60: Unused Variables (aka stupid stupid compiler...)
@bencouturier found that the default gcc5 compiler on Ubuntu throws an `unused-variable` warning for the clock tick variables in `utils.h`. Then combined with `-Werror` the build fails.

This looks to me like a buggy compiler, because no other compiler we test against (gcc4.7, gcc6.2, gcc7.3) shows this issue.

I had a whirl at patching this with the `#pragma GCC diagnostic ignored "-Wunused-variable"`, however, it seems that this is quite buggy in `g++` (problems observed up to at least gcc6.3, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53431).

As noted, *stupid stupid compiler...*

The use of these clock tick variables is probably only marginally useful, so maybe it's best just to use `sysconf(_SC_CLK_TCK)` directly where needed.

graeme-a-stewart:
Some gcc compiler versions claim these are unused variables and throw a compiler warning, that bombs the build with the -Werror option being set.

This fixes #60.

### DIFF
--- a/package/src/cpumon.cpp
+++ b/package/src/cpumon.cpp
@@ -38,7 +38,7 @@ void cpumon::update_stats(const std::vector<pid_t>& pids) {
     }
     stat_entries.clear();
   }
-  for (auto& stat : cpu_stats) stat.second *= prmon::inv_clock_ticks;
+  for (auto& stat : cpu_stats) stat.second /= sysconf(_SC_CLK_TCK);
 }
 
 // Return the summed counters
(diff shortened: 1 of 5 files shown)


graeme-a-stewart:
It might be an error of compilation unit scope. Later compilers maybe realise that a `static` is used as long as it's used anywhere in the code. The weird version of gcc seems to be considering its use only in a specific compilation unit. But the error only seems to pop up for "derived" variables - the ones set to pure constants it treats correctly.

Thanks for the merge.

PR #62, opened 2018-05-31
Profiling options

### LINKED ISSUE #58: Support a profile build option
`prmon` resource consumption seems to be very reasonable, even for 8 core multi-process jobs (0.5% as seen in #50), however we should support a profiling build option so that we can analyse where we do actually consume resources.

graeme-a-stewart:
I worked on how to enable profiling and concluded that it's really not worthwhile generically supporting this at the level of CMake (very few people ever will need to do this). So instead I added some notes on how to make a small change in the `CMakeLists.txt` that enables either gprof or gperftools.

This is good enough to fix #58.

I also improved a few points in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ prmon currently runs on Linux machines as it requires access to the
 ### Building the project
 
 Building prmon requires a modern C++ compiler, CMake version 3.1 or
-higher and the RapidJSON libraries. Note that the installation of
+higher and the [RapidJSON libraries](http://rapidjson.org/). Note that the installation of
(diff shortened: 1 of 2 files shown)


graeme-a-stewart:
[pprof2901.0.pdf](https://github.com/HSF/prmon/files/2057500/pprof2901.0.pdf)

This is the gperftools output from running prmon aggressively over an ATLAS 8-core q431 job. It demonstrates what I expected - the vast majority of time is spend reading from the smaps files for the athena child processes.

PR #63, opened 2018-05-31
Add a CMake build option to generate a static binary and support profiling

graeme-a-stewart:
For a utility like prmon having a statically build binary is a very useful thing to then be able to ship it around between machines and OSs, so add the correct compiler and linker options to do that.

This was suggested by @bencouturier - thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,16 +20,37 @@ include(GNUInstallDirs)
 if(NOT CMAKE_CONFIGURATION_TYPES)
   if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE RelWithDebInfo
-      CACHE STRING "Choose the type of build, options are: None Release MinSizeRel Debug RelWithDebInfo"
+      CACHE STRING "Choose the type of build, options are: None Release MinSizeRel Debug RelWithDebInfo Static"
       FORCE
       )
   else()
     set(CMAKE_BUILD_TYPE "${CMAKE_BUILD_TYPE}"
-      CACHE STRING "Choose the type of build, options are: None Release MinSizeRel Debug RelWithDebInfo"
+      CACHE STRING "Choose the type of build, options are: None Release MinSizeRel Debug RelWithDebInfo Static"
       FORCE
       )
   endif()
 endif()
+
+# - Define the static build type as this is very useful for a utility like prmon
+SET( CMAKE_CXX_FLAGS_STATIC "-O2" CACHE STRING
+    "Flags used by the C++ compiler during static builds."
+    FORCE )
+SET( CMAKE_C_FLAGS_STATIC "-O2" CACHE STRING
+    "Flags used by the C compiler during static builds."
+    FORCE )
+SET( CMAKE_EXE_LINKER_FLAGS_STATIC
+    "-Wl,--warn-unresolved-symbols,--warn-once -static" CACHE STRING
+    "Flags used for linking binaries during static builds."
+    FORCE )
+SET( CMAKE_SHARED_LINKER_FLAGS_STATIC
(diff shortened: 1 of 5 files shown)


graeme-a-stewart:
Hmmm, seems that this also builds the test binaries statically, which means they consume different amounts of virtual memory and the vmem check fails in `singleMem` and `childMem`.

That needs to be fixed in some smart way.

graeme-a-stewart:
OK, all done properly now and tested, ready for review,

graeme-a-stewart:
Thanks @amete - well spotted indeed. I need to switch to something like `find_package(Gperftools REQUIRED)` instead. Then that also allows people to set some steering variables for cases where the tools are in a non-standard location.

bcouturi:
I tried the branch on Ubuntu 16.04, with the system compiler, with the following options:

```
cmake -DBUILD_STATIC=true -DRapidJSON_INCLUDE_DIR=/opt/rapidjson-1.1.0/include -DCMAKE_INSTALL_PREFIX=/opt/prmon ../prmon
```
and the static binary produced shows the following behaviour:

```
$ /opt/prmon/bin/prmon
One and only one PID or child program is required - found none
Segmentation fault (core dumped)
$
```

There seems to be a segmentation fault when returning from the main method.
I will try on other platforms for comparison.

graeme-a-stewart:
Hi @bcouturi - I tried and I get a clean exit on CentOS7, but I can reproduce the core dump on Ubuntu 16.04. However the segfault is baffling as all that's done it to print an `endl` then call `return` (see  https://github.com/HSF/prmon/blob/2bec960e5c578b50d44c8eefee6c919465f8376f/package/src/prmon.cpp#L320).

```sh
Core was generated by `./package/prmon'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00000000004d5808 in __run_exit_handlers ()
#2  0x00000000004d5853 in exit ()
#3  0x00000000004c439d in generic_start_main ()
#4  0x00000000004c458a in __libc_start_main ()
#5  0x0000000000402709 in _start ()
```

bcouturi:
Indeed, very puzzling. I'm trying to check that all the objects used in the binary are compiled correctly.

bcouturi:
Shall we just keep a note and investigate at a later stage ?

graeme-a-stewart:
Hmmm, if I remove the threading library (`pthreads`) and all of the gymnastics with condition variables then the Debian static binary seems to work fine.

Reviewing the use of the condition variable, it only seems to provide the functionality that when `prmon` gets a `SIGUSR1` then it will exit immediately, breaking any sleep. However, if we were to reduce the sleep time a bit (like 250ms) then we could still exit on `SIGUSR1`, with a maximum delay of that time. 

I'm really not sure who actually calls `SIGUSR1` for `prmon`, unless it's the ATLAS pilot? Or was it just a feature that was implemented, but never actually used?

graeme-a-stewart:
OK, I got ride of all the condition variable stuff in the last commit. That simplifies things and AFAICS now the static binary on Ubuntu 16.04 works fine.

graeme-a-stewart:
I now fixed @amete's original issue. There is a FindGperftools script added and it's possible to hint to CMake using Gperftools_ROOT_DIR. So, in the SLC6 case do:

```sh
cmake -DPROFILE_GPERFTOOLS=ON -DGperftools_ROOT_DIR=/cvmfs/atlas-nightlies.cern.ch/repo/sw/21.0/sw/lcg/releases/LCG_88/gperftools/2.5/x86_64-slc6-gcc62-opt PATH/TO/prmon/
```

That works for me.

Finally, this is GTG!

bcouturi:
All fine for me !

Issue #65, opened 2018-06-06
Tweaks to axis labels and plot titles needed


sciaba:
For prmon plots to be good for publications and presentations, some minor issues with overlapping text labels should be fixed. See as an example:

![prmon](https://user-images.githubusercontent.com/12764525/41041101-7a551fb0-699e-11e8-84f2-4918a0ddd2de.png)

amete:
I just pushed a new branch named `master-improve-plotting-script` where I tried to address this issue. I tried to beautify the axis/legend labels, properly assign the units etc. This also addresses the plotting issues mentioned in #64.

Some examples are:

`./prmon_plot.py --xvar wtime --yvar vmem,pss,rss --stacked`
![prmon_wtime_vs_vmem_pss_rss](https://user-images.githubusercontent.com/7428938/41043737-de3731ac-69a4-11e8-88bb-a729746e7903.png)

`./prmon_plot.py --xvar wtime --yvar utime,stime,wtime`
![prmon_wtime_vs_utime_stime_wtime](https://user-images.githubusercontent.com/7428938/41043745-e4d33f7e-69a4-11e8-8c62-794501e2bd57.png)

`./prmon_plot.py --xvar wtime --yvar utime,stime --diff`
![prmon_wtime_vs_diff_utime_stime](https://user-images.githubusercontent.com/7428938/41043755-eb287e84-69a4-11e8-9f88-52d7a6248f11.png)

`./prmon_plot.py --xvar wtime --yvar rx_bytes,tx_bytes --diff`
![prmon_wtime_vs_diff_rx_bytes_tx_bytes](https://user-images.githubusercontent.com/7428938/41043764-f02a9570-69a4-11e8-96eb-6569c8336e28.png)

Could you give it a shot and let me know? Let me tag @graeme-a-stewart in case he also wants to comment.

Many thanks.

-s

graeme-a-stewart:
@amete - thanks! My suggestion is that one can somehow scale the y-axis into a fixed unit size, e.g., GB or kB/s. When I see memory in "kB * 1e8" I have to do quite some mental gymnastics to realise that 0.2 = 20GB.

BTW, I like a lot plotting the time in relative seconds instead of the actual GMT clock time - much more useful.

amete:
[...]

Issue #64, opened 2018-06-06
Adding time derivatives of some metrics


sciaba:
Hi,
I would humbly suggest to add time derivatives of some metrics, namely the IO and the network metrics. As we know, spikes in IO rates are extremely important when characterising an application and having prmon_plot.py being able to generate plots for them would be very handy.
IMO it can be done just in prmon_plot.py, no need to touch prmon.
Cheers,
Andrea

amete:
Hi Andrea,

If I understand your request correctly the core functionality should be there already, i.e.:

```
prmon_plot.py --input ... --xvar wtime --yvar rx_bytes
```

would create the usual time evolution of `rx_bytes` like:

![prmon_wtime_vs_rx_bytes](https://user-images.githubusercontent.com/7428938/41039829-edef094e-699a-11e8-84d1-f9818cea5289.png)

while

```
prmon_plot.py --input ... --xvar wtime --yvar rx_bytes --diff
```

would compute and plot delta(yvar)/delta(xvar) as:

![prmon_wtime_vs_diff_rx_bytes](https://user-images.githubusercontent.com/7428938/41039890-1de7b952-699b-11e8-9b40-4d2783b8379a.png)

I think we can improve the labeling etc. but is this what you're after?

Best,
Serhan

sciaba:
Exactly! The only degree of freedom is the delta_t used to estimate the derivative. It should be a multiple of the sampling interval; some degree of smoothing may be desirable (maybe a sensible default and a configurable parameter? Up to you!).
Cheers,
Andrea

sciaba:
I was thinking again, and maybe it's not just a matter of plots, but we should consider IO rates as "real" metrics, which means having their values written already in prmon.txt, such that prmon.json will contain max and avg of them, too... Clearly it's more work because they should be calculated in prmon itself.
Andrea

graeme-a-stewart:
Hi @sciaba - I am not so sure about that. From the accumulating values it's easy to calculate the derivative and we could easily adopt a few strategies to do that (various sliding window shapes and widths). If you take the derivative inside prmon and then only give rates it's not so easy to go backwards.

Technically it's not *that* hard to do, but I what we do now is more useful and reflects the underlying metric better, so it's easier to relate to other measurements.

sciaba:
You convinced me!

amete:
[...]

PR #66, opened 2018-06-06
Various improvements to the plotting script

amete:
In this PR we improve the plots in various ways. Labels and units are set more uniformly. The user now has the ability to specify the units of the plotted variables and the format of the output image. Fixes the visual/plotting issues discussed in #64 and #65.

Tagging @sciaba for further comments in case there is any.

### DIFF
--- a/package/scripts/prmon_plot.py
+++ b/package/scripts/prmon_plot.py
@@ -18,6 +18,54 @@ except ImportError:
     print('Please install these modules first and then retry. ')
     sys.exit(-1)
 
+# Define the labels/units for beautification
+axisunits = {'vmem':'kb', 'pss':'kb', 'rss':'kb',
+             'utime':'s', 'stime':'s', 'wtime':'s',
+             'rchar':'char','wchar':'char',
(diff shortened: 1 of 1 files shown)


package/scripts/prmon_plot.py
```diff
@@ -18,21 +19,90 @@
...
+               'read_bytes':'I/O Read',
+               'write_bytes':'I/O Written',
+               'rx_packets':'Network Received',
+               'tx_packets':'Network Transmitted',
+               'rx_bytes':'Network Received',
+               'tx_bytes':'Network Transmitted'}
+
+multipliers = {'SEC': 1.,
+               'MIN': 60.,
+               'HR': 60.*60.,
```
graeme-a-stewart: I would change the name here to `HOUR` - having a plot labled with units of `HR` is a bit unclear to me.

amete:
Thanks a lot @graeme-a-stewart. I just changed `HR` to `HOUR` as you suggested.

graeme-a-stewart:
Fixes #65.

PR #68, opened 2018-06-07
Add swap to the plotting script and update the readme

amete:
I just noticed that I was missing the `swap` in the `prmon_plot.py`, which is now added. I extended the README to include a bit more information about the capabilities of the plotting script, and added three example plots along w/ the associated commands. The plots are hosted in the repo itself, something I'm not sure what others would think. I think it might be OK but I'm open to suggestions of course.


amete:
Thanks a lot @graeme-a-stewart. I added the file in the same folder as the plots and mentioned this in the README.

Issue #69, opened 2018-06-08
Lack of platform name handled badly


graeme-a-stewart:
Trying to compile prmon on a gentoo linux system (part of the HSF Packaging group's test drive) I got this error:

```
CMake Error at cmake/prmonCPack.cmake:93 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.
Call Stack (most recent call first):
  cmake/prmonCPack.cmake:136 (hsf_get_platform)
  CMakeLists.txt:69 (include)
```

This seems to come from an empty/undefined value for `HSF_OS_VERSION`.

So this should be handled better for sure.

(There might well be an upstream issue for the HSF package template creator.)

amete:
Fixed in #75, closing.

PR #75, opened 2018-06-12
Handle correctly situation where /etc/os-release is incomplete

graeme-a-stewart:
If the parsing of /etc/os-release failed to set either HSF_OS_ID or
HSF_OS_VERSION there was then a nasty CMake error thrown
in a later REGEX REPLACE. At least for bleeding edge releases
the VERSION_ID string can be absent (noticed on Gentoo portage)
so this broke.

In these cases we set the strings to some value that's consistent with
not knowing the correct value ("unknown v0").


graeme-a-stewart:
Fixes #69

PR #76, opened 2018-06-12
Preparing for v1.0.1

amete:
Once in master, I'll merge master into stable and tag from stable. Here we set the correct version for the package version in the CMake file.


graeme-a-stewart:
N.B. `stable` can be fast forwarded, so please don't create a merge commit!

PR #85, opened 2018-08-23
Updating Travis configuration and README

amete:
I simply turned on the push builds on travis, updated the yaml configuration such that we only build master/stable branches, and added some information to the README regarding the build status and the license.  I believe when we merge this to the master branch it should invoke a build whose result should then appear on the front page.

Next I'll take a look at possibility of adding coverity stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,15 @@
 sudo: required
 
+# Add only master and stable branches
+branches:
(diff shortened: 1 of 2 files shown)


amete:
BTW, I'm realizing that .com and .org are different on Travis somehow.

PR #92, opened 2018-11-29
Switch from RapidJSON to nlohmann/json

### LINKED ISSUE #88: Consider use of nlohmann/json code
There was some discussion on ATLAS lists that concluded that people favoured [Niels Lohmann's JSON code for C++](https://github.com/nlohmann/json). It's a header only library that can actually be made as concise as a single (albeit large) file to be included. 

It would be good to evaluate that as an alternative to RapidJSON, especially if it could be provided easily bundled with prmon.

amete:
Hi @graeme-a-stewart,

Here is the PR as we briefly discussed. This should fix #88. Please let me know, many thanks.

Best,
Serhan

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,7 @@ prmon currently runs on Linux machines as it requires access to the
 ### Building the project
 
 Building prmon requires a modern C++ compiler, CMake version 3.1 or
-higher and the [RapidJSON libraries](http://rapidjson.org/). Note that the installation of
-RapidJSON needs to be modern enough that CMake is supported (e.g.,
-on Ubuntu 16.04 `rapidjson-dev` is too old, just install it yourself).
+higher and the [nlohmann\_json libraries](https://github.com/nlohmann/json). 
 
 Building should be as simple as
 
@@ -38,8 +36,8 @@ Building should be as simple as
     make -j<number of cores on your machine>
     make install
 
-If your installation of RapidJSON is in a non-standard location then
-setting `RapidJSON_DIR` may be required as a hint to CMake.
+If your installation of nlohmann\_json is in a non-standard location then
+setting `nlohmann\_json\_DIR` may be required as a hint to CMake.
 
 The option `-DCMAKE_BUILD_TYPE` can switch between all of the standard
 build types. The default is `Release`; use `RelWithDebInfo` if you want
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@
 
 [build-img]: https://travis-ci.com/HSF/prmon.svg?branch=master
 [build-link]: https://travis-ci.com/HSF/prmon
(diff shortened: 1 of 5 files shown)


amete:
I believe the build fails since we don't have `nlohmann/json` in the containers. We can re-build after fixing that.

package/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ add_executable(prmon src/prmon.cpp
 					 src/countmon.cpp 
 					 src/wallmon.cpp
 					 src/memmon.cpp)
-target_include_directories(prmon PRIVATE ${RapidJSON_INCLUDE_DIR})
+target_link_libraries(prmon PRIVATE nlohmann_json::nlohmann_json)
```
graeme-a-stewart: Oddly, this line causes a failure when building on Ubuntu18 with the native installation of Nlohmann's JSON (`nlohmann-json-dev`).

Removing it seems to work find and not prejudice the CentOS7 build, when the library is taken from EPEL (`json-devel`).

This might be because the Ubuntu18 version is surprisingly old, 2.1.1, vs. the C7 3.0.0.

README.md
```diff
@@ -38,8 +36,8 @@ Building should be as simple as
     make -j<number of cores on your machine>
     make install
 
-If your installation of RapidJSON is in a non-standard location then
-setting `RapidJSON_DIR` may be required as a hint to CMake.
+If your installation of nlohmann\_json is in a non-standard location then
```
graeme-a-stewart: Why the `\`?

graeme-a-stewart:
I fixed the c7-dev container to have the nlohmann/json library in it, so that now works fine (https://travis-ci.com/HSF/prmon/jobs/161733235).

I started work on the Ubuntu18 container and ran into the issue with the `target_link_libraries` line noted above. Also strange is that I need to manually tell Ubuntu's CMake where the support files are:

```
cmake -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_C_COMPILER=$CC -Dnlohmann_json_DIR=/usr/lib/cmake /mnt/code/prmon
```

I will need to adjust the Travis build script to account for this.

graeme-a-stewart:
OK, this is all looking good. I updated the test containers for C7 and U16. There is a glitch in the U18 container that at the moment I can't fix as one of their default repositories is throwing errors when applying security updates. Just need to wait for them to fix that and then the U18 test container can be fixed and updated.

graeme-a-stewart:
Hi @amete - so I went back to this and rebuilt that Ubuntu 18 container. All the tests pass now, so I think this is GTG.

graeme-a-stewart:
But as I made some commits on the PR, you should probably sign off on them @amete.

amete:
All looks fantastic, thanks a lot @graeme-a-stewart!

Issue #97, opened 2019-03-03
Monitoring frequency?


sciaba:
Would it be feasible to add the possibility to monitor the frequency of the cores on which the application is running, or of all the cores on the system (easier and probably more sensible)?
A simple

cat /proc/cpuinfo | grep MHz

shows in real time the actual frequency of each core. This would allow to find out if the CPU was using a turbo frequency or was throttling at any point during the execution of the application.

Cheers,
Andrea

amete:
Hi @sciaba,

The only issue that I foresee here is that we need to somehow now the proc/thread <-> CPU association. This would need some extra bookkeeping which might be a bit convoluted. We can simply dump the /proc/cpuinfo but then we don't really know if the scaling potentially affect the process that we're listening. We also probably need to store the baseline frequency (min/max) of the processor to judge how much scaling is going one. I'll give this some thought but please let me know if you have any suggestions. Also let me tag @graeme-a-stewart in case he has an idea.

Many thanks.

Best,
Serhan

sciaba:
Ideally what you say is correct, but I think it would be acceptable just to record the absolute clock speed of each core in the system. This is what any motherboard monitor tool would do, it's not imperative to identify the cores used by the process(es), also because they may well change over time and the time charts would be difficult to interpret. A first implementation along these lines should be trivial and would allow users to understand if it's sufficient.

Cheers,
Andrea

amete:
Thanks Andrea. If we're only after the information whether or not scaling is enabled, I think there is a neater solution. On Linux CPUfreq exposes CPU information here:

`/sys/devices/system/cpu/cpu{CPUID}/cpufreq/scaling_governor`

where the `CPUID` is essentially the index in 0...Nprocessors-1. The scaling is off if this reads `performance` and on otherwise (in most recent versions the only other option is `powersave`). We can read this over all CPUs and store a single column w/ a boolean in the text file for each measurement. Would this suffice you think?

-s

sciaba:
Hi Serhan,
I would still like to see a clock frequency... maybe you are right, but the clock frequency would allow e.g. to compare plots obtained on different CPUs which may have different base and boost clock frequencies.

Cheers,
Andrea

amete:
Hi @graeme-a-stewart,

Apparently I took a stab at this in #102 but we didn't resolve it. How would you like to proceed? Please let me know, many thanks.

Best,
Serhan

graeme-a-stewart:
[...]

PR #102, opened 2019-06-06
Adding CPU frequency information

amete:
This PR addresses #97. The executable now accepts a boolean, `--store-cpu-freq` (or `-s` in short), which turns on processor frequency information collection. If enabled (off by default), we have new columns in the text file that read `CPUX` and the clock freq at the time of sampling is stored in MHz. The JSON file will have the average and peak values as well.

Regardless of the above, we now store an additional column named `cpu_scaling` by default in the text file. This holds information about whether or not the CPU scaling is on. We store 0 for false, 1 for true and 2 in case something goes wrong in the collection. The JSON file only contains the peak value as the average doesn't make much sense for this guy.

Needlessly to say I tested everything locally and all checks out. Given that this is a fairly substantial update I'd appreciate a careful check/comments/suggestions. I'm explicitly cc-ing @sciaba in case he wants to test this earlier.

### DIFF
--- a/package/src/cpuinfo.cpp
+++ b/package/src/cpuinfo.cpp
@@ -44,8 +44,8 @@ int cpuinfo::get_number_of_cpus() {
   return nCPUs;
 }
 
-// Method to get the processor clock speeds
-std::vector<float> cpuinfo::get_processor_clock_speeds() {
+// Method to get the processor clock freqs
+std::vector<float> cpuinfo::get_processor_clock_freqs() {
   std::vector<float> result;
 
   std::ifstream cpuInfoFile{"/proc/cpuinfo"};
@@ -63,7 +63,7 @@ std::vector<float> cpuinfo::get_processor_clock_speeds() {
     if (splitIdx != std::string::npos) val = line.substr(splitIdx + 1);
     if (line.size() >= key.size() && line.compare(0, key.size(), key) == 0) {
        result.push_back(std::stof(val));
-    } // end of filling clock speeds - order doesn't change 
+    } // end of filling clock freqs - order doesn't change
   } // end of reading cpuInfoFile  
 
   cpuInfoFile.close();
@@ -72,8 +72,9 @@ std::vector<float> cpuinfo::get_processor_clock_speeds() {
 }
 
 // Method to get the scaling information
-int cpuinfo::cpu_scaling_info(int nCPUs) {
-  bool result = false;
+// 0 for false, 1 for true, 2 for error
+unsigned int cpuinfo::cpu_scaling_info(int nCPUs) {
+  unsigned int result = 0;
 
   for(int cpu = 0; cpu < nCPUs; ++cpu) {
     std::string fileName = "/sys/devices/system/cpu/cpu" + std::to_string(cpu) + "/cpufreq/scaling_governor";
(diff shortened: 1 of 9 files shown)


package/src/prmon.cpp
```diff
@@ -219,26 +219,29 @@ int main(int argc, char* argv[]) {
...
       {"json-summary", required_argument, NULL, 'j'},
       {"interval", required_argument, NULL, 'i'},
+      {"store-cpu-freq", no_argument, NULL, 's'},
       {"netdev", required_argument, NULL, 'n'},
       {"help", no_argument, NULL, 'h'},
       {0, 0, 0, 0}};
 
   char c;
-  while ((c = getopt_long(argc, argv, "p:f:j:i:n:h", long_options, NULL)) !=
+  while ((c = getopt_long(argc, argv, "p:f:j:i:s:n:h", long_options, NULL)) !=
```
graeme-a-stewart, line 244: This should just be `s` and not `s:`, which indicates that the short option should take an argument

package/src/cpuinfo.cpp
```diff
@@ -0,0 +1,96 @@
...
+// 0 for false, 1 for true, 2 for error
+unsigned int cpuinfo::cpu_scaling_info(int nCPUs) {
+  unsigned int result = 0;
+
+  for(int cpu = 0; cpu < nCPUs; ++cpu) {
+    std::string fileName = "/sys/devices/system/cpu/cpu" + std::to_string(cpu) + "/cpufreq/scaling_governor";
+
+    std::ifstream governorFile{fileName};
+    if(!governorFile.is_open()) {
+      std::cerr << "Failed to open " << fileName << std::endl;
```
graeme-a-stewart, line 84: In my C7 docker container on the Mac this file doesn't exist, so I get an error message printed every iteration - I guess if some kernels don't have this we should be quiet here (OTOH, `/proc/cpuinfo` should always be there)

amete:
I'll let @sciaba to comment on the usefulness of this since he requested the functionality. As I originally commented in #97 `prmon` mainly targets to monitor the resource usage associated to a given process. Just listening to all the CPUs and dumping their clock frequencies w/o any association between the procs/threads is more on the hardware monitoring side than the process monitoring side. Although the this can be resolved to some extend (w/ possibly quite a bit of bookkeeping gymnastics), I also think this is a bit out of the scope of what `prmon` wants to accomplish. I wouldn't mind scrapping the PRs completely.

amete:
BTW, from a technical point of view, the new `cpuinfo` can become its own plug-in and we can remove the knobs from `cpumon` completely. Therefore, the technical comments are easy to address but I guess the most important thing is to converge on whether we really want this in `prmon` or not.

graeme-a-stewart:
> BTW, from a technical point of view, the new `cpuinfo` can become its own plug-in and we can remove the knobs from `cpumon` completely. Therefore, the technical comments are easy to address but I guess the most important thing is to converge on whether we really want this in `prmon` or not.

Yes, I do agree with that

sciaba:
I agree that this feature request can be dropped. Different cores may run at different speeds, processes and threads can be moved to different cores at any time, so the feature is indeed ill defined.
My initial thought was to be able to recognise if the CPU was throttling or being turbo-boosted, but in modern CPUs clock speeds may change on a time scale of milliseconds, so (unless we are in the pathological case of severe throttling due to bad cooling) clock variations simply can happen all the time.

PR #109, opened 2020-05-27
Streamline JSON object use

### LINKED ISSUE #108: Support averages as floats
As noted in #105 we have a rounding issue when the values that prmon reports are small integers. This happens in the calculation of average values in JSON stats, where `nprocs`, `nthreads` and `ngpus` are reported.

Fixing this requires changing the signature of the average values from `unsigned long long` to `float` (thus in `Imonitor.h` as well as all of the concrete monitoring classes).

graeme-a-stewart:
When starting to fix #108 I had a chance to review the way we handle the JSON summary object in prmon. It seems there was a lot of cruft left over, possibly from the previous use of the rapidjson implemenetation. This PR removes some of that anachronistic code and optimises handling.

Additionally, I renamed output file names to be more descriptive.

### DIFF
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -61,7 +61,7 @@ void SignalChildHandler(int /*signal*/) {
 }
 
 int MemoryMonitor(const pid_t mpid, const std::string filename,
-                  const std::string jsonSummary, const unsigned int interval,
+                  const std::string json_summary_file, const unsigned int interval,
                   const std::vector<std::string> netdevs) {
   signal(SIGUSR1, SignalCallbackHandler);
   signal(SIGCHLD, SignalChildHandler);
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
Ooops, I think I might have broken the C7 docker image used by Travis

amete:
By the way, @graeme-a-stewart, do you want to fix the docker image for **c7-dev** and rerun the CI before merging #109 and #110? In both cases **u18-dev** seems to have passed, so I think I can simply merge them w/ confidence but let me know.

graeme-a-stewart:
I think I should be able to retrigger the CI now that the C7 test is fixed...

graeme-a-stewart:
Hmmm, hold the door, this still seems to use the devtoolset-7 version, even though master was updated with the new configuration

graeme-a-stewart:
OK, I tried restarting the whole Travis job, not just the sub-job that failed

graeme-a-stewart:
Of course, it makes sense - the `.travis.yml` is from before the fix... Rebased

PR #110, opened 2020-05-27
Implement average values as doubles

### LINKED ISSUE #108: Support averages as floats

graeme-a-stewart:
Having average value calculations as unsigned integers was producing harsh
rounding errors for small valued quantities (like nprocs, nthreads). These are
now implemented as double precision values and get serialised as
numbers in JSON with decimal fractions.

There is a slight drawback that the decimal part is designed in nlohmann::json
to avoid precision loss, so sometimes a very high number of DPs are printed.
However, there doesn't seem to be any way to adjust this without manually
massaging the actual numerical values (there are various issues in
https://github.com/nlohmann/json where this is discussed).

An example is:

```
a3febc5a3b43 [hsf]:prmon-build$ cat prmon.json 
{
    "Avg": {
        "nprocs": 1.0,
        "nthreads": 5.0,
        "pss": 1813.0,
        "rchar": 846.1538461538462,
        "read_bytes": 0.0,
        "rss": 3528.0,
        "rx_bytes": 0.0,
        "rx_packets": 0.0,
        "swap": 0.0,
        "tx_bytes": 0.0,
        "tx_packets": 0.0,
        "vmem": 55876.0,
        "wchar": 8.333333333333334,
        "write_bytes": 0.0
    },
    "Max": {
        "nprocs": 1,
        "nthreads": 5,
        "pss": 1813,
        "rchar": 5280,
        "read_bytes": 0,
        "rss": 3528,
        "rx_bytes": 0,
        "rx_packets": 0,
        "stime": 0,
        "swap": 0,
        "tx_bytes": 0,
        "tx_packets": 0,
        "utime": 24,
        "vmem": 55876,
        "wchar": 52,
        "write_bytes": 0,
        "wtime": 6
    }
}
```

Fixes #108

### DIFF
--- a/package/src/Imonitor.h
+++ b/package/src/Imonitor.h
@@ -23,7 +23,7 @@ class Imonitor {
 
   virtual std::map<std::string, unsigned long long> const get_text_stats() = 0;
   virtual std::map<std::string, unsigned long long> const get_json_total_stats() = 0;
-  virtual std::map<std::string, unsigned long long> const get_json_average_stats(unsigned long long elapsed_clock_ticks) = 0;
+  virtual std::map<std::string, double> const get_json_average_stats(unsigned long long elapsed_clock_ticks) = 0;
 
 };
 
--- a/package/src/countmon.cpp
+++ b/package/src/countmon.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) CERN, 2018
+// Copyright (C) CERN, 2020
 
 #include "countmon.h"
 #include "utils.h"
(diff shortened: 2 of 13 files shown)


graeme-a-stewart:
Yes, it's completely good enough, even if the number of DPs is OTT. It's a very slight irritation that the file is larger than it really needs to be, but we can live with it...

Issue #107, opened 2020-05-25
Options to enable/disable monitoring components


graeme-a-stewart:
Currently `prmon` monitors basically everything it can. The #105 issue is adding support for NVIDIA GPU monitoring, which will only happen after a few tests.

This issue collects opinions on if we should allow prmon components to be enabled/disabled on the command line. At the moment there has been no request to do this and possibly no one really wants it...

amete:
To be honest, I would implement this if there is a specific request from the users. For the time being the number of metrics we collect is fairly reasonable and there is not much overhead due to collecting all. If this changes in the future it might indeed be good to support toggling different components on/off but I don't see the need at this point of time from my experience.

graeme-a-stewart:
So, implementing #113 actually made this quite strightforward to implement, so although there was no specific request for it, it seemed doable and useful. Implemeneted in #118.

Issue #106, opened 2020-05-20
Add logging functionality


graeme-a-stewart:
When developing the nvidia monitoring in #105 I was constantly adding and removing debug `std::cout` statements to try and understand/debug the code. Some of these are definitely throwaway, but it really would be useful to be able to switch on logging for `prmon`. e.g., I can imagine with GPU monitoring that what we implement might not work on all platforms and we'd need some debug output from a user to understand why.

In the spirit of not reinventing the wheel, I searched for lightweight logging libraries for C++ and there are a few, notably:

- https://github.com/SergiusTheBest/plog
- https://github.com/gabime/spdlog

Which are header only, which I think is attractive.

I notice that `spdlog` is in EPEL and it seems to be more popular, so I'd probably try this one first.

PR #115, opened 2020-06-02
Plugin registry implementation

### LINKED ISSUE #113: Add registry for monitoring components
Thinking about #106 and #107 these could be nicely implemented by introducing a registry solution in prmon, where every monitoring component gets registered, allowing the main prmon binary to query the register for components. This would allow

- Each component to carry some simple information that could be printed as part of the `--help` message
- Each component to have a "name" that can be used for enabling/disabling (#107) and for setting a log level (#106) via the CLI

This should be implementable as registry class in a header file (there are quite a lot of examples floating around on stackoverflow and github).

graeme-a-stewart:
This is not quite finished in it's full implementation, but it is a good time for review as further development comes on top.

Inspired by <https://github.com/psalvaggio/cppregpattern> this is a registry implementaion for prmon.

It makes the code in the main monitor look a lot nicer, as initialisation is now a loop over the registry. I am not completely happy with the pointer grabbed to `wallmon`, but as it's used inside the monitoring loop (to pass wallclock time to the average stats generation) it is needed and I could not think of a better way.

Closes #113 
Helps with making #106 and #107 much easier to implement

### DIFF
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -10,12 +10,10 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
-#include <condition_variable>
 #include <cstddef>
 #include <fstream>
 #include <iostream>
-#include <map>
-#include <mutex>
+#include <unordered_map>
 #include <sstream>
 #include <thread>
 #include <vector>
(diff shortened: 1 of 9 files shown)


package/src/prmon.cpp
```diff
@@ -67,31 +65,30 @@ int MemoryMonitor(const pid_t mpid, const std::string filename,
...
+      }
+      monitors.push_back(std::move(new_monitor_p));
+    } else {
+      std::cerr << "Registration of monitor " << class_name << " FAILED" << std::endl;
+    }
+  }
+  if (wallclock_monitor_p == nullptr) {
+    std::cerr << "Failed to initialise wallclock monitoring class" << std::endl;
+    exit(EXIT_FAILURE);
+  }
```
amete, line 91: I agree, this bit for the `wallmon` is not so nice. I wonder if we can implement a `get` method in the factory that'll return the singleton based on the `class_name`. That way, perhaps these bits might be tied up a bit.

graeme-a-stewart: I don't favour the idea that the registry itself keeps track of the instantiated classes - singletons are just a bit fragile, so I prefer the fact that the registry is basically compile time constructed. 

Here, keeping track of what prmon is using is the job of the `monitors` vector of unique pointers. Of course instead of a vector we could make that an `unordered_map` so that we could have the name of the component as a key. There's a penalty in iterating over a map instead of a vector - maybe irrelevant though... It would be a more consistent design to support this 'feature'.

graeme-a-stewart: Yep, I think it's a lot better with the change to a map (https://github.com/HSF/prmon/pull/115/commits/466faad930782e5d1745ff259ca7dbd9145a3a61)

graeme-a-stewart:
Hi @amete - so I think this in good shape to me merged now, if you approve of the changes

amete:
Hi @graeme-a-stewart,  Do we really want to return a raw pointer w/ `registry::Registry<Imonitor>::create`? In the main application, we feed this into a `std::unique_ptr` so I don't think there is a problem per se, but I thought it would be nicer if the factory dealt w/ `std::unique_ptr` instead of raw pointers. If you're happy w/ this, I'll go ahead and merge.

graeme-a-stewart:
That's a good point! Yes, let me try to make that change then.

amete:
All looks good to me, merging now.

PR #116, opened 2020-06-03
Hardware information implementation

### LINKED ISSUE #114: Add hardware information to the JSON file
Let me create this issue so that it is not forgotten. The idea is to add some hardware information on the CPU and the GPU (if it exists) into the JSON file. We can simply create a new section, say `hw`, and populate it accordingly.

I think we should agree on what we want to store. On the CPU side we typically have easy access to the following information (also available through `/proc/cpuinfo`):

```
$ lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                32
On-line CPU(s) list:   0-31
Thread(s) per core:    2
Core(s) per socket:    8
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 62
Model name:            Intel(R) Xeon(R) CPU E5-2667 v2 @ 3.30GHz
Stepping:              4
CPU MHz:               1200.036
CPU max MHz:           4000.0000
CPU min MHz:           1200.0000
BogoMIPS:              6623.89
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              25600K
NUMA node0 CPU(s):     0-7,16-23
NUMA node1 CPU(s):     8-15,24-31
Flags:                 fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aperfmperf eagerfpu pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm epb ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid fsgsbase smep erms xsaveopt dtherm ida arat pln pts md_clear spec_ctrl intel_stibp flush_l1d
``` 

and on the GPU side, since we're currently only supporting NVIDIA we can simply use `nvidia-smi -L` which gives, e.g.:

```
$ nvidia-smi -L
GPU 0: GeForce RTX 2060 (UUID: GPU-XXX)
```

I guess at the bare minimum we want to keep the number of CPUs/GPUs along w/ their model names.

As to how we'd like to do this, I'm completely impartial. One option is to do this directly in `prmon.cpp` outside the monitoring loop. We can either write helper functions to collect this information directly there, or collect them inside the relative plug-ins (e.g. a la `nvidiamon::test_nvidia_smi()`) and pass them on to the main application as we do for the actual metrics (although this might be a bit awkward if we want to collect the hardware information regardless of whether we run the relevant plug-in).

amete:
This is still WIP but I think the basic idea is there. Therefore, it might be good to have a discussion before expanding on it.

Now the main interface has a `get_hardware_info()` method that is implemented in each plug-in to collect and return some hardware information. A basic implementation is done in `cpumon` to populate the chip vendor name as well as the number of processors that are defined. 

The collection is off by default and can be enabled w/ an additional argument, `--store-hw-info`, on the `prmon` side.  This results in the creating of a new section in the JSON file, named `HW`, and looks something like this at the moment:

```

  "Avg": {
    "nprocs": 1.0,
    "nthreads": 0.0,
    "pss": 122.0,
    "rchar": 2123.9583333333335,
    "read_bytes": 0.0,
    "rss": 548.0,
    "rx_bytes": 302.6041666666667,
    "rx_packets": 1.5625,
    "swap": 0.0,
    "tx_bytes": 647.1354166666666,
    "tx_packets": 2.6041666666666665,
    "vmem": 4372.0,
    "wchar": 0.0,
    "write_bytes": 0.0 
  },  
  "HW": {
    "cpu": {
      "model name": " Intel(R) Xeon(R) CPU E5-2667 v2 @ 3.30GHz",
      "nCPU": "32"
    }   
  },  
  "Max": {
    "nprocs": 1,
    "nthreads": 0,
    "pss": 122,
    "rchar": 8156,
    "read_bytes": 0,
    "rss": 548,
    "rx_bytes": 1162,
    "rx_packets": 6,
    "stime": 0,
    "swap": 0,
    "tx_bytes": 2485,
    "tx_packets": 10, 
    "utime": 0,
    "vmem": 4372,
    "wchar": 0,
    "write_bytes": 0,
    "wtime": 3
  }
}
```

I think we should have a discussion on what metrics we want to collect for in which plug-ins. The usual `/proc` exposes a number of things that we collect. I personally would like to stay away from using command-line utilities (especially those that require root privileges) but GPU information will possibly be an exception to that.

In any case, this is still a prototype but a decent starting point IMO.

Closes #114.

### DIFF
--- a/package/src/Imonitor.h
+++ b/package/src/Imonitor.h
@@ -15,6 +15,8 @@
 #include <string>
 #include <vector>
 
+#include <nlohmann/json.hpp>
+
 class Imonitor {
  public:
   virtual ~Imonitor() {};
@@ -25,7 +27,7 @@ class Imonitor {
   virtual std::map<std::string, unsigned long long> const get_json_total_stats() = 0;
   virtual std::map<std::string, double> const get_json_average_stats(unsigned long long elapsed_clock_ticks) = 0;
 
-  virtual std::map<std::string, std::map<std::string, std::string>> const get_hardware_info() = 0;
+  virtual void const get_hardware_info(nlohmann::json& j) = 0;
 };
 
 #endif  // PRMON_IMONITOR_H
--- a/package/src/Imonitor.h
+++ b/package/src/Imonitor.h
@@ -27,7 +27,7 @@ class Imonitor {
   virtual std::map<std::string, unsigned long long> const get_json_total_stats() = 0;
   virtual std::map<std::string, double> const get_json_average_stats(unsigned long long elapsed_clock_ticks) = 0;
 
-  virtual void const get_hardware_info(nlohmann::json& j) = 0;
+  virtual void const get_hardware_info(nlohmann::json& hw_json) = 0;
 };
 
 #endif  // PRMON_IMONITOR_H
--- a/package/src/Imonitor.h
+++ b/package/src/Imonitor.h
@@ -25,6 +25,7 @@ class Imonitor {
   virtual std::map<std::string, unsigned long long> const get_json_total_stats() = 0;
   virtual std::map<std::string, double> const get_json_average_stats(unsigned long long elapsed_clock_ticks) = 0;
 
+  virtual std::map<std::string, std::map<std::string, std::string>> const get_hardware_info() = 0;
 };
 
 #endif  // PRMON_IMONITOR_H
--- a/package/src/cpumon.cpp
+++ b/package/src/cpumon.cpp
@@ -71,8 +71,8 @@ std::map<std::string, std::map<std::string, std::string>> const cpumon::get_hard
   }
 
   // Metrics to read from the input
-  std::vector<std::string> metrics{"processor", "model name"};
-  unsigned int nCPU = 0;
+  std::vector<std::string> metrics{"processor", "model name", "siblings", "cpu cores"};
+  unsigned int nCPU = 0, nSiblings = 0, nCores = 0;
 
   // Loop over the file
   std::string line;
@@ -86,6 +86,8 @@ std::map<std::string, std::map<std::string, std::string>> const cpumon::get_hard
       for (const auto& metric : metrics) {
         if (line.size() >= metric.size() && line.compare(0, metric.size(), metric) == 0) {
           if (metric == "processor") nCPU++;
+          else if (metric == "siblings") { if(nSiblings == 0) nSiblings = std::stoi(val); }
+          else if (metric == "cpu cores") { if(nCores == 0) nCores = std::stoi(val); }
           else if (result["cpu"][metric].empty()) {
             result["cpu"][metric] = std::regex_replace(val, std::regex("^\\s+|\\s+$"), "");
           }
@@ -94,8 +96,15 @@ std::map<std::string, std::map<std::string, std::string>> const cpumon::get_hard
     } // end of seperator check
   } // end of reading cpuInfoFile
 
-  // Fill nCPU
+  // The clear assumption here is that there is a single type of processor installed
+  // nCPU = nSockets * nSiblings
+  // nSiblings = nCores * nThreads
+  unsigned int nSockets = nCPU / nSiblings;
+  unsigned int nThreads = nSiblings / nCores;
   result["cpu"]["nCPU"] = std::to_string(nCPU);
+  result["cpu"]["nSockets"] = std::to_string(nSockets);
+  result["cpu"]["nCores"] = std::to_string(nCores);
+  result["cpu"]["nThreads"] = std::to_string(nThreads);
 
   // Close the file
   cpuInfoFile.close();
--- a/package/src/cpumon.cpp
+++ b/package/src/cpumon.cpp
@@ -9,6 +9,7 @@
 #include <fstream>
 #include <iostream>
 #include <sstream>
+#include <regex>
 
 // Constructor; uses RAII pattern to be valid
 // after construction
@@ -62,7 +63,7 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
 std::map<std::string, std::map<std::string, std::string>> const cpumon::get_hardware_info() {
   std::map<std::string, std::map<std::string, std::string>> result{};
 
-  // Read some information from /proc/cpuinfo 
+  // Read some information from /proc/cpuinfo
   std::ifstream cpuInfoFile{"/proc/cpuinfo"};
   if(!cpuInfoFile.is_open()) {
     std::cerr << "Failed to open /proc/cpuinfo" << std::endl;
@@ -79,17 +80,19 @@ std::map<std::string, std::map<std::string, std::string>> const cpumon::get_hard
     if (line.empty()) continue;
     size_t splitIdx = line.find(":");
     std::string val;
-    if (splitIdx != std::string::npos) val = line.substr(splitIdx + 1);
-    for (const auto& metric : metrics) {
-      if (line.size() >= metric.size() && line.compare(0, metric.size(), metric) == 0) {
-        if (val.empty()) continue;
-        if (metric == "processor") nCPU++;
-        else if (result["cpu"][metric].empty()) {
-          result["cpu"][metric] = val;
-        }    
-      }
-    } // end of populating metrics
-  } // end of reading cpuInfoFile  
+    if (splitIdx != std::string::npos) {
+      val = line.substr(splitIdx + 1);
+      if (val.empty()) continue;
+      for (const auto& metric : metrics) {
+        if (line.size() >= metric.size() && line.compare(0, metric.size(), metric) == 0) {
+          if (metric == "processor") nCPU++;
+          else if (result["cpu"][metric].empty()) {
+            result["cpu"][metric] = std::regex_replace(val, std::regex("^\\s+|\\s+$"), "");
+          }
+        } // end of metric check
+      } // end of populating metrics
+    } // end of seperator check
+  } // end of reading cpuInfoFile
 
   // Fill nCPU
   result["cpu"]["nCPU"] = std::to_string(nCPU);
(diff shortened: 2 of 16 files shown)


package/src/cpumon.cpp
```diff
@@ -57,3 +57,45 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
...
+  std::vector<std::string> metrics{"processor", "model name"};
+  unsigned int nCPU = 0;
+
+  // Loop over the file
+  std::string line;
+  while (std::getline(cpuInfoFile,line)) {
+    if (line.empty()) continue;
+    size_t splitIdx = line.find(":");
+    std::string val;
+    if (splitIdx != std::string::npos) val = line.substr(splitIdx + 1);
```
graeme-a-stewart, line 82: If we don't find the separator, shouldn't we also now skip the rest of the loop? e.g. put the metric test loop inside this `if`:

```
if (splitIdx != std::string::npos) {
  val = line.substr(splitIdx + 1);
  for (const auto& metric : metrics) {
    ...
  }
}
```

(or use `continue`)

package/src/prmon.cpp
```diff
@@ -260,6 +270,8 @@ int main(int argc, char* argv[]) {
         << default_json_summary << ")\n"
         << "[--interval, -i TIME]     Seconds between samples (default "
         << default_interval << ")\n"
+        << "[--store-hw-info, -s]     Store hardware information (default "
+        << default_store_hw_info << ")\n"
```
graeme-a-stewart, line 274: This prints a `0` which isn't really nice. How about changing to
```
<< default_store_hw_info ? "true" : "false" << ...
```

package/src/memmon.cpp
```diff
@@ -75,3 +75,9 @@ std::map<std::string, double> const memmon::get_json_average_stats(
     unsigned long long elapsed_clock_ticks) {
   return mem_average_stats;
 }
+
+// Collect related hardware information
+std::map<std::string, std::map<std::string, std::string>> const memmon::get_hardware_info() {
+  std::map<std::string, std::map<std::string, std::string>> result{};
+  return result;
```
graeme-a-stewart, line 82: It would be nice here to parse `/proc/meminfo` and get `MemTotal`

amete:
Hi @graeme-a-stewart,

Thanks a lot. I tidied things up a bit based on your comments. We now report the total available memory and I added some extra CPU information. The output now looks something like this:

```
{
  "Avg": {
    "nprocs": 1.0,
    "nthreads": 0.0,
    "pss": 190.0,
    "rchar": 523.9583333333334,
    "read_bytes": 0.0,
    "rss": 616.0,
    "rx_bytes": 1146.6145833333333,
    "rx_packets": 6.510416666666667,
    "swap": 0.0,
    "tx_bytes": 1371.3541666666667,
    "tx_packets": 7.03125,
    "vmem": 4432.0,
    "wchar": 0.0,
    "write_bytes": 0.0 
  },  
  "HW": {
    "cpu": {
      "model name": "Intel(R) Xeon(R) CPU E5-2667 v2 @ 3.30GHz",
      "nCPU": "32",
      "nCoresPerSocket": "8",
      "nSockets": "2",
      "nThreadsPerCore": "2" 
    },  
    "mem": {
      "MemTotal": "32723876 kB"
    }   
  },  
  "Max": {
    "nprocs": 1,
    "nthreads": 0,
    "pss": 190,
    "rchar": 2012,
    "read_bytes": 0,
    "rss": 616,
    "rx_bytes": 4403,
    "rx_packets": 25, 
    "stime": 0,
    "swap": 0,
    "tx_bytes": 5266,
    "tx_packets": 27, 
    "utime": 0,
    "vmem": 4432,
    "wchar": 0,
    "write_bytes": 0,
    "wtime": 3
  }
}
```

For the extra CPU information, I added some extra information as to how I compute them (essentially `nCPU = nSockets * nCoresPerSocket * nThreadsPerCore`). A check would be nice but a least on my machine `lscpu` reports consistent results:

```
CPU(s):                32
Thread(s) per core:    2
Core(s) per socket:    8
Socket(s):             2
```

Let me know how it looks.

package/src/cpumon.cpp
amete: That's correct, I reorganized this bit.

package/src/prmon.cpp
amete: Good point, updated.

package/src/memmon.cpp
amete: Good point, now added.

package/src/cpumon.cpp
```diff
@@ -57,3 +58,56 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
...
+
+  // The clear assumption here is that there is a single type of processor installed
+  // nCPU = nSockets * nSiblings
+  // nSiblings = nCoresPerSocket * nThreadsPerCore
+  unsigned int nSockets = nCPU / nSiblings;
+  unsigned int nThreads = nSiblings / nCores;
+  result["cpu"]["nCPU"] = std::to_string(nCPU);
+  result["cpu"]["nSockets"] = std::to_string(nSockets);
+  result["cpu"]["nCoresPerSocket"] = std::to_string(nCores);
+  result["cpu"]["nThreadsPerCore"] = std::to_string(nThreads);
```
graeme-a-stewart, line 107: I don't like that these values com out as strings, that seems unnatural. So thinking about this maybe the underlying storage object should be a JSON entity, instead of the more rigid map of strings. That would allow for a more natural representation of this information.

If we were re-mapping post-facto then we'd have to somehow communicate the intrinsic type back to the main program, which seems complicated and not necessary if we went the JSON route. As this information is gathered only once then performance isn't a concern.

package/src/memmon.cpp
```diff
@@ -75,3 +76,40 @@ std::map<std::string, double> const memmon::get_json_average_stats(
...
+  while (std::getline(memInfoFile,line)) {
+    if (line.empty()) continue;
+    size_t splitIdx = line.find(":");
+    std::string val;
+    if (splitIdx != std::string::npos) {
+      val = line.substr(splitIdx + 1);
+      if (val.empty()) continue;
+      for (const auto& metric : metrics) {
+        if (line.size() >= metric.size() && line.compare(0, metric.size(), metric) == 0) {
+          result["mem"][metric] = std::regex_replace(val, std::regex("^\\s+|\\s+$"), "");
```
graeme-a-stewart, line 105: I would report this without the `kB` units in it (consistency with the way the rest of prmon works, Taking into account the above comment, it should be a JSON number, not a string.

amete:
Hi @graeme-a-stewart,

Thanks a lot. I updated the interface a bit. Now instead of returning a map, we pass the `json` object by reference and fill it inside the `get_hardware_info` method. This way we're more flexible in terms of types, so currently we have:

```
{
  "Avg": {
    "nprocs": 1.0,
    "nthreads": 0.0,
    "pss": 186.0,
    "rchar": 2123.9583333333335,
    "read_bytes": 0.0,
    "rss": 612.0,
    "rx_bytes": 557.03125,
    "rx_packets": 3.125,
    "swap": 0.0,
    "tx_bytes": 1105.46875,
    "tx_packets": 4.166666666666667,
    "vmem": 4432.0,
    "wchar": 0.0,
    "write_bytes": 0.0 
  },  
  "HW": {
    "cpu": {
      "model name": "Intel(R) Xeon(R) CPU E5-2667 v2 @ 3.30GHz",
      "nCPU": 32, 
      "nCoresPerSocket": 8,
      "nSockets": 2,
      "nThreadsPerCore": 2
    },  
    "mem": {
      "MemTotal": 32723876
    }   
  },  
  "Max": {
    "nprocs": 1,
    "nthreads": 0,
    "pss": 186,
    "rchar": 8156,
    "read_bytes": 0,
    "rss": 612,
    "rx_bytes": 2139,
    "rx_packets": 12, 
    "stime": 0,
    "swap": 0,
    "tx_bytes": 4245,
    "tx_packets": 16, 
    "utime": 0,
    "vmem": 4432,
    "wchar": 0,
    "write_bytes": 0,
    "wtime": 3
  }
}
```

Now what I don't like too much is that the `nlohmann/json.hpp` is included in the interface becase forward declaring this guy is a bit awkward. The library needs to be built w/ a special flag `-DJSON_MultipleHeaders=ON` which also gives an additional `fwd` header but I couldn't get it to work quite nicely. I'm not sure if this is a showstopper for you but at least it seems to be working.

Let me know what you think.

package/src/memmon.cpp
amete: Done.

package/src/cpumon.cpp
amete: Done (see https://github.com/HSF/prmon/pull/116#issuecomment-638838785 for more details).

package/src/Imonitor.h
```diff
@@ -25,6 +27,7 @@ class Imonitor {
   virtual std::map<std::string, unsigned long long> const get_json_total_stats() = 0;
   virtual std::map<std::string, double> const get_json_average_stats(unsigned long long elapsed_clock_ticks) = 0;
 
+  virtual void const get_hardware_info(nlohmann::json& j) = 0;
```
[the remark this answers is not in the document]
graeme-a-stewart: (and for all the concrete implmentations as well)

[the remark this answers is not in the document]
amete: Indeed, `j` is fairly obscure, switched to `hw_json` throughout.

amete:
I just removed the WIP, I think this is ready to go whenever you're happy w/ it @graeme-a-stewart. Thanks a lot 😄

PR #123, opened 2020-06-04
Update to Python3 and add new CI targets

### LINKED ISSUE #44: netBurner.py only works in python2
As noted in #37 netBurner.py only works in python2. This is because of the significant restructuring of the `urllib2` library between py2 and py3.

We should see if this can be fixed with the help of the `six` module.

### LINKED ISSUE #83: Put test docker containers into HSF dockerhub area
As noted by @amete in #82 the docker containers used to test prmon should be moved into the HSF dockerhub area.

### LINKED ISSUE #122: Python3
As noticed in #112 we have an issue that newer Linux platforms usually only come with Python3 by default these days (Python2 is officially out of support from 2020-01-01).

As noted in <https://www.python.org/dev/peps/pep-0394/>, on unix systems `python3` and `python2` are preferred to refer to major Python versions. `python` is a wildcard - you could get either one (or nothing at all!).

I propose that we bite the bullet and make the next version of `prmon` Python3. Note that:

- Python is only needed for tests and for the plotter code, `prmon` itself is obviously compiled code and depends on CMake and the compiler.
- Python3 was even available on very old distributions (lxplus6 has Python3.4; Ubuntu16 has Python3.5).

So all scripts should update to `python3` shebang.

Note that the very old issue #44 can now also be solved properly.

One open question would be the minimum version to support. Format strings only arrived with Python3.6, but we may feel that dropping support for SLC6 and Ubuntu16 is still premature.

graeme-a-stewart:
Scripts to test networking are more extensively changed
as the Python3 modules are different from those in Python2

For the CI, Ubuntu20 and CentOS8 are added

Closes #122 
Closes #44 

Note the Python3 status in the various CI platforms:

| Distro | Python3 |
|-----|-----|
| Ubuntu16 | 3.5 |
| CentOS7 | 3.6 |
| Ubuntu18 | 3.6 |
| CentOS8 | 3.6 |
| Ubuntu20 | 3.8 |

Also updated to HSF docker containers

Closes #83


.travis.yml
```diff
@@ -15,6 +15,8 @@ env:
   - DIMAGE=graemeastewart/u16-dev
```
amete, line 15: I think we can drop u16 as per our discussion at #122.

.travis.yml
```diff
@@ -30,3 +32,5 @@ script:
   - if [ "$DIMAGE" = graemeastewart/u16-dev ]; then docker run -e CXX=g++-7 -e CC=gcc-7 -v $(pwd):/mnt $DIMAGE /mnt/.travis-scripts/build-test.sh; fi
```
amete, line 32: I think we can drop u16 as per our discussion at #122.

package/tests/CMakeLists.txt
```diff
@@ -51,11 +51,11 @@ add_test(NAME basicIOmultithread COMMAND testIO.py --usleep 100 --io 10 --thread
 add_test(NAME basicIOmultiproc COMMAND testIO.py --usleep 100 --io 10  --procs 2)
 
 # Net Tests
-add_test(NAME basicNET COMMAND python testNET.py)
+add_test(NAME basicNET COMMAND python3 testNET.py)
 
 # Memory Tests
-add_test(NAME singleMem COMMAND python testMEM.py --procs 1)
-add_test(NAME childMem COMMAND python testMEM.py --procs 4)
+add_test(NAME singleMem COMMAND python3 testMEM.py --procs 1)
```
amete, line 57: I never really noticed this before. Is there a specific reason why some of our tests (e.g. this one) are `... COMMAND pythonX foo.py ...` and others (e.g. CPU ones) `... COMMAND foo.py ....`? Perhaps we can unify them.

graeme-a-stewart: Interesting - I am really not sure why this is... see comment below

graeme-a-stewart:
So, regarding the test invokation, it makes me think, what about we invoke using an *explicit* Python, set as a CMake variable?

```
add_test(NAME singleMem COMMAND PYTHON testMEM.py --procs 1)
```

Then we can pretty easily switch the test setup from Python3 to Python2, even if we say the default Python is v3 (seom the shebang). For the network test we do need to switch between two different scripts, but surely that's manageable?

graeme-a-stewart:
Oh yep, so I tried building this branch in `Athena,master,latest`. The `prmon` build is just fine, but the tests fall over hard:

```sh
lxplus773:~/tmp/build$ make test
Running tests...
Test project /afs/cern.ch/user/g/graemes/tmp/build
      Start  1: testCPUsingle
 1/11 Test  #1: testCPUsingle ....................Child aborted***Exception:   0.29 sec
...

0% tests passed, 11 tests failed out of 11
```

So `python3` is completely screwed up here (which also means that *detecting* Python3 from the presence of a `python3` binary doesn't at all guarrantee that it would work).

amete:
I think what really screw things up is overwriting PYTHONPATH in asetup (which happens when setting up a nightly). In principle, when we're building the projects we simply setup gcc and cmake, i.e. `lsetup 'asetup none,gcc8,cmakesetup --cmakeversion=3.17.3'`. In that case, PYTHONPATH seems not defined, so testing might still be OK at that stage.

graeme-a-stewart:
It's true - if I unset PYTHONPATH and PYTHONHOME then `python3` is functional again.

.travis.yml
graeme-a-stewart: Done in latest commits

graeme-a-stewart:
OK, I think I have a good solution now. By default Python3 is used, but the version of Python used in the tests can be set by the user to be something else via the `PYTHON_TEST` option for CMake. So to go back to Python2:

```
cmake -DPYTHON_TEST=python2 ...
```
(It's also possible to set an absolute path here.)

The tests will run different versions of the network test in `python2` and `python3` setups because the Python library modules for http changed significantly between versions. (If someone sets just `python` then we skip the test as it's not possible to guess which one would be correct; invoking `python` and checking the version seems like too much, given this is just a transitional work-around).

Building prmon with `master,latest,Athena` now works and passes all tests when setting the above `-DPYTHON_TEST=python2` option.

We have to keep the suite of test scripts (excepting the network test) compatible with both Pythons for a while. To be sure we don't break any of the test scripts for Python2 there is a new CI option to test in Python2 on CentOS7.

graeme-a-stewart:
Last commit changed to using HSF versions of the docker containers

Closes #83

PR #124, opened 2020-06-05
Remove the offset in nthreads

### LINKED ISSUE #121: The number of threads that is reported by countmon
Now when we're counting the number of processes and threads in `countmon`, we offset the mother thread from number of threads, i.e.:

https://github.com/HSF/prmon/blob/5af921b643971300b287d30680318c9383d5ec81/package/src/countmon.cpp#L40-L41

The rationale behind this was to keep track of threads that are explicitly created in each process but I don't think we ever discussed/agreed on this choice. What do you think @graeme-a-stewart? Shall we leave as it's or simply report what we read?

amete:
As discussed in #121, for each process we were subtracting 1 from `nthreads` to offset the mother thread from accounting. Although there is no loss of information, this might be confusing and is against the common pattern of simply reporting what we measure. Therefore, this PR gets rid of it. The test script is also updated to take this change into account.

Closes #121.

### DIFF
--- a/package/tests/testCOUNT.py
+++ b/package/tests/testCOUNT.py
@@ -45,7 +45,7 @@ def setupConfigurableTest(threads=1, procs=1, time=10.0, interval=1, invoke=Fals
                 totPROC      = prmonJSON["Max"]["nprocs"] 
                 totTHREAD    = prmonJSON["Max"]["nthreads"]
                 expectPROC   = procs
-                expectTHREAD = procs*threads 
+                expectTHREAD = procs*(threads+1)
                 self.assertAlmostEqual(totPROC, expectPROC, msg = "Inconsistent value for number of processes "
                              "(expected {0}, got {1})".format(expectPROC, totPROC))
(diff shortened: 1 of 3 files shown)


package/tests/testCOUNT.py
```diff
@@ -45,7 +45,7 @@ def test_runTestWithParams(self):
                 totPROC      = prmonJSON["Max"]["nprocs"] 
                 totTHREAD    = prmonJSON["Max"]["nthreads"]
                 expectPROC   = procs
-                expectTHREAD = procs*threads 
+                expectTHREAD = procs*(threads+1)
```
graeme-a-stewart, line 48: I think the issue here is in the burner, which is actually written so that it *spawns* an additional `n` threads, leaving one parent thread that doesn't run a CPU burner, but just waits for its worker threads to finish (hence then `n+1` total threads).

I think it's more sensible to fix the burner to really use `n` threads, instead of `n+1`.

PR #125, opened 2020-06-11
Add NVIDIA GPU monitoring

### LINKED ISSUE #105: Add optional GPU monitoring support
Hi,

I'm creating this issue to follow-up on the discussion that was held at the ATLAS SPOT meeting where we discussed adding GPU monitoring support to `prmon`.

Currently the easiest and the most reliable/useful information seems to be provided by the utilities that are developed by the hardware manufacturers, e.g. `nvidia-smi`.  

For the initial implementation, the agreed-upon idea is to write a new plug-in, where we invoke `nvidia-smi` to gather GPU statistics and add these to the rest. The rationale behind this choice is the wide usage of NVIDIA accelerators where `prmon` is used, the minimum dependency at the compilation/installation time (as opposed to using the C API) etc. The implementation will be done with the understanding that support for other hardware, eg. Intel/AMD, can/will be added in the future.

Looking at the relevant documentation, e.g. [here](https://developer.download.nvidia.com/compute/DCGM/docs/nvidia-smi-367.38.pdf), it seems as if `nvidia-smi` supports different options, the most relevant ones being `dmon` which provides device monitoring and `pmon` which provides process monitoring statistics. There are a number of concerns in terms of the overhead of using the command-line utility etc. but these will need to be assessed as the project progresses.

Please feel free to provide relevant input under this ticket. Many thanks.

Best,
Serhan

graeme-a-stewart:
Closes #105 

This code adds a monitor for NVIDA GPUs to prmon. When prmon starts it tries to invoke `nvidia-smi` to see if there are relevant GPUs to monitor. If it finds any then monitoring is enabled.

The statistics printed are:

- `gpusmpct` - sum of the streaming multiprocessor usage from this job as a %age value (thus 89 means 89% use, and so on; can be >100% when multiple GPUs are active)
- `gpumempct` - sum of the memory bandwidth used from this job as a %age value
- `gpufbmem` - sum of fb memory, in **kB**
- `ngpus` - number of GPUs actively used by this job

The list of processes from the monitored job is used to ensure that statistics are only gathered for this job family (so if other users are utilising the GPU, this is ignored).

If hardware statistics (`--store-hw-info`) are requested then the following will be reported:

```
        "gpu": {
            "gpu_0": {
                "name": "TITAN Xp",
                "sm_freq": 1911,
                "total_mem": 12488704
            },
            "gpu_1": {
                "name": "TITAN Xp",
                "sm_freq": 1911,
                "total_mem": 12487680
            },
            "nGPU": 2
        },
```

i.e., the total number of GPUs found, then for each GPU the model name, the clock frequency of the streaming multiprocessors (MHz) and the total amount of memory (kB).

There are some general utility enhancements included in this PR:

1. A new utility function, `cmd_pipe_output` is added (helper for running external commands)
1. `prmon` no longer sets a general `SIGCHLD` handler, as it is fragile when being used in combination with forked children in `cmd_pipe_output`
    1. The main `prmon` loop simply reaps child processes' exit codes every cycle

As the `nvidiamon` code needs to check if it has valid things to monitor, an `is_valid()` method is added to the `Imonitor` virtual base class. All other monitors simple return `true`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 cmake_minimum_required(VERSION 3.3)
 # - Call project() to setup system
 # From CMake 3, we can set the project version easily in one go
-project(prmon VERSION 1.1.1)
+project(prmon VERSION 1.2.0)
 
 #--- Define basic build settings -----------------------------------------------
 # - Use GNU-style hierarchy for installing build products
--- a/package/CMakeLists.txt
+++ b/package/CMakeLists.txt
@@ -3,12 +3,12 @@ find_package(nlohmann_json REQUIRED)
 add_executable(prmon src/prmon.cpp
            src/prmonutils.cpp
            src/utils.cpp
-					 src/netmon.cpp
-					 src/iomon.cpp
-					 src/cpumon.cpp
-					 src/countmon.cpp
-					 src/wallmon.cpp
-					 src/memmon.cpp
+           src/netmon.cpp
+           src/iomon.cpp
+           src/cpumon.cpp
+           src/countmon.cpp
+           src/wallmon.cpp
+           src/memmon.cpp
            src/nvidiamon.cpp
            )
 
--- a/package/CMakeLists.txt
+++ b/package/CMakeLists.txt
@@ -3,7 +3,6 @@ find_package(nlohmann_json REQUIRED)
 add_executable(prmon src/prmon.cpp
            src/prmonutils.cpp
            src/utils.cpp
-					 src/prmonutils.cpp
 					 src/netmon.cpp
 					 src/iomon.cpp
 					 src/cpumon.cpp
--- a/package/CMakeLists.txt
+++ b/package/CMakeLists.txt
@@ -3,12 +3,15 @@ find_package(nlohmann_json REQUIRED)
 add_executable(prmon src/prmon.cpp
            src/prmonutils.cpp
            src/utils.cpp
-           src/netmon.cpp
-           src/iomon.cpp
-           src/cpumon.cpp
-           src/countmon.cpp
-           src/wallmon.cpp
-           src/memmon.cpp)
+					 src/prmonutils.cpp
+					 src/netmon.cpp
+					 src/iomon.cpp
+					 src/cpumon.cpp
+					 src/countmon.cpp
+					 src/wallmon.cpp
+					 src/memmon.cpp
+           src/nvidiamon.cpp
+           )
 
 # Some older versions of the nlohmann_json package don't define
 # this target (< 3.0.0 ?). The build on these platforms seems to
--- a/package/src/nvidiamon.cpp
+++ b/package/src/nvidiamon.cpp
@@ -1,0 +1,194 @@
+// Copyright (C) CERN, 2020
+
+#include "nvidiamon.h"
+#include "utils.h"
+
+#include <string>
+#include <iostream>
+#include <sstream>
+
+#include <unistd.h>
+#include <stdlib.h>
+
+// This is not very nice, but we don't want to add more special runtime
+// flags, pending a proper decision on logging (#106)
+#define NVIDIA_DEBUG 0
+
+// Constructor; uses RAII pattern to be valid after construction
+nvidiamon::nvidiamon() : nvidia_params{prmon::default_nvidia_params},  
+  nvidia_stats{}, nvidia_average_stats{}, nvidia_total_stats{}, iterations{0L} {
+  for (const auto& nvidia_param : nvidia_params) { 
+    nvidia_stats[nvidia_param] = 0;
+    nvidia_peak_stats[nvidia_param] = 0;
+    nvidia_average_stats[nvidia_param] = 0;
+    nvidia_total_stats[nvidia_param] = 0;
+  }
+
+  // Attempt to execute nvidia-smi
+  // If this works we are valid, but if not then we can't get data
+  valid = test_nvidia_smi();
+}
+
+void nvidiamon::update_stats(const std::vector<pid_t>& pids) {
+  const std::vector<std::string> cmd = {"nvidia-smi", "pmon", "-s", "um", "-c", "1"};
+
+  for (auto& stat : nvidia_stats) stat.second = 0;
+
+  auto cmd_result = prmon::cmd_pipe_output(cmd);
+  if (cmd_result.first) {
+    // Failed
+    return;
+  }
+
+  if (NVIDIA_DEBUG) {
+    std::cout << "nvidiamon::update_stats got the following output (" << cmd_result.second.size() << "): " << std::endl;
+    int i = 0;
+    for (const auto& s : cmd_result.second) {
+      std::cout << i << " -> " << s << std::endl;
+      ++i;
+    }
+  }
+
+  // Loop over output
+  unsigned int gpu_idx, sm, mem, enc, dec, fb_mem;
+  pid_t pid;
+  std::string cg_type, cmd_name;
+  std::vector<unsigned int> activegpus{}; // Avoid double counting active GPUs
+  for (const auto& s : cmd_result.second) {
+    if (s[0] != '#') {
+      std::istringstream instr(s);
+      instr >> gpu_idx >> pid >> cg_type >> sm >> mem >> enc >> dec >> fb_mem >> cmd_name;
+      if (!(instr.fail() || instr.bad())) {
+        if (NVIDIA_DEBUG) {
+          std::cout << "Good read: " << gpu_idx << " " << pid << " " << 
+            cg_type << " " << sm << " " << mem << " " << enc << " " << 
+            dec << " " << fb_mem << " " << cmd_name << std::endl;
+        }
+
+        // Filter on PID value, so we only add stats for our processes
+        for (auto const p : pids) {
+          if (p == pid) {
+            nvidia_stats["gpusmpct"] += sm;
+            nvidia_stats["gpumempct"] += mem;
+            nvidia_stats["gpufbmem"] += fb_mem;
+            bool new_gpu = true;
+            for (auto const gpu : activegpus) {
+              if (gpu == gpu_idx) new_gpu = false;
+            }
+            if (new_gpu) {
+              activegpus.push_back(gpu_idx);
+              ++nvidia_stats["ngpus"];
+            }
+          }
+        }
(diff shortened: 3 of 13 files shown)


graeme-a-stewart:
P.S. If problems are experienced with the NVIDIA monitor, setting `NVIDIA_DEBUG` to `1` in `nvidiamon.cpp` will enable some useful debugging output. We have only been able to test this code on the few machines we have available to us, with their GPU cards and nvidia-smi setups.

graeme-a-stewart:
Large rebase, plus squash and applied clang-format fixes...

graeme-a-stewart:
So, @amete, we got a fail mark from CodeFactor as it thinks `nvidiamon::update_stats` is too complex. Any idea how we see exactly what the failed metric is? Does it just that it thinks there are too many lines of code for a single method?

amete:
I think so, it's probably complaining because it's a large block of code. Actually if you click on the "Details" next to CodeFactor check, hover your mouse over the "Complex Method" header an arrow will pop-up on the right that reads "Read up" (when you're over it). When you click on it it give a bit more information (albeit generic). Maybe it's because we're using the free version but I cannot seem to be able to find an easy way to configure this. Therefore, for the time-being I removed it from the "Required" list so that it doesn't block the PR. I think it's good to have this check but perhaps we don't want it to be too restrictive.

graeme-a-stewart:
Yep, I actually found that button and clicking on it does gives some tips.

I had a think about it and I found a few ways to reduce the complexity, using different data structures and getting rid of multi-condition if statements... so on the next commit let's see I managed to reduce the number of code paths (it was `complexity = 19`).

amete:
@graeme-a-stewart, actually bugging @matthewfeickert, one of the core developers of `scikit-hep/pyhf`, I think we can add a top level `codecov.yml` file and configure some of the settings there. From a quick search, I see the defaults are [here](https://github.com/codefactor-io/default-configs/blob/master/detekt.yml). The default threshold seems to be 0 which is fairly stringent. So, perhaps something along the lines of:

```
complexity:
  active: true
  ComplexMethod:
    active: true
    threshold: N
```

with a meaningful `N` would be a good trade-off. It seems as if we have a few main functions that trigger this. We can also turn this specific check off, too.

matthewfeickert:
Yeah for CodeFactor I don't believe there is a config file API (check me on this) so you will need to login to the account an in the website just ignore things manually or set them as account wide ignores. For coverage stuff though CodeCov is a great way to go. I'm happy to discuss any questions on this too.

graeme-a-stewart:
[CodeFactor](https://www.codefactor.io/repository/github/hsf/prmon) found an issue: Complex Method

It's currently on:
[package\src\nvidiamon.cpp:37-123
](https://github.com/HSF/prmon/blob/8cec9c94730d64e66142dd68f9c306165a7d36aa/package/src/nvidiamon.cpp#L37-L123)Commit 8cec9c94730d64e66142dd68f9c306165a7d36aa

graeme-a-stewart:
OK, so I a bit puzzled - I did change that code, seemingly for the better:

https://github.com/HSF/prmon/pull/125/commits/5a8f0c3287330523f9d9af26cb6f4b4d303a8134?file-filters%5B%5D=.cpp

However, CodeFactor is still showing an issue, **for the old code**, and failing our check. Doesn't it update on new commits? What am I missing?

matthewfeickert:
> Doesn't it update on new commits? What am I missing?

Yes, but it might be stuck. I'd just toggle it or ignore it. Worth reporting these things too as the dev team actually does respond on the forums.

amete:
Yes, seems like it's stuck on the previous push. Perhaps you can try a dummy change and push again @graeme-a-stewart?

I'm a bit disappointed that things are not configurable. Oh well. Indeed all the configuration seems to go through the account page as descried [here](https://support.codefactor.io/i18-ignoring-issues-and-patterns). Looks like it allows ignoring specific files, specific issues or types of issues but they all have to be done through the dashboard.

package/src/nvidiamon.cpp
```diff
@@ -0,0 +1,207 @@
...
+    if (nvidia_stats[nvidia_param] > nvidia_peak_stats[nvidia_param]) {
+      nvidia_peak_stats[nvidia_param] = nvidia_stats[nvidia_param];
+    }
+    nvidia_total_stats[nvidia_param] += nvidia_stats[nvidia_param];
+    nvidia_average_stats[nvidia_param] =
+        double(nvidia_total_stats[nvidia_param]) / iterations;
+  }
+}
+
+// Return the nvidiaers
```
amete, line 121: Is there a typo here?

package/src/nvidiamon.cpp
```diff
@@ -0,0 +1,207 @@
...
+  auto cmd_result = prmon::cmd_pipe_output(cmd);
+  if (cmd_result.first != 0) return false;
+  unsigned int gpus = 0;
+  for (auto const& s : cmd_result.second) {
+    // From C++20 can use 'starts_with'
+    if (s.substr(0, 3).compare("GPU") == 0) {
+      ++gpus;
+    }
+  }
+  nvidiamon::ngpus = gpus;
```
amete, line 154: I think we can introduce a basic `set` method and set the members that way. I think it would be nicer.

package/src/nvidiamon.cpp
```diff
@@ -0,0 +1,207 @@
...
+                 "unreliable"
+              << std::endl;
+  }
+  return true;
+}
+
+// Collect related hardware information
+void const nvidiamon::get_hardware_info(nlohmann::json& hw_json) {
+  // Record the number of GPUs
+  hw_json["HW"]["gpu"]["nGPU"] = nvidiamon::ngpus;
```
amete, line 169: Getting this w/ a constant `get` method would be nicer.

package/src/nvidiamon.cpp
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  // Loop over output
+  unsigned int gpu_idx{}, sm{}, mem{}, enc{}, dec{}, fb_mem{};
+  pid_t pid{};
+  std::string cg_type{}, cmd_name{};
+  std::unordered_map<unsigned int, bool>
+      activegpus{};  // Avoid double counting active GPUs
+  for (const auto& s : cmd_result.second) {
+    if (s[0] != '#') continue;
```
amete, line 67: I think this is supposed to be:

```suggestion
    if (s[0] == '#') continue;
```

as it stands now I'm getting all zeros.

package/src/nvidiamon.cpp
```diff
@@ -0,0 +1,207 @@
...
+                  << " " << sm << " " << mem << " " << enc << " " << dec << " "
+                  << fb_mem << " " << cmd_name << std::endl;
+      }
+
+      // Filter on PID value, so we only add stats for our processes
+      for (auto const p : pids) {
+        if (p == pid) {
+          nvidia_stats["gpusmpct"] += sm;
+          nvidia_stats["gpumempct"] += mem;
+          nvidia_stats["gpufbmem"] += fb_mem;
```
amete, line 84: It seems as if we're still accumulating memory in MB, we probably need:

```suggestion
          nvidia_stats["gpufbmem"] += fb_mem * 1024;
```

to convert this into KB to be consistent.

CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 cmake_minimum_required(VERSION 3.3)
 # - Call project() to setup system
 # From CMake 3, we can set the project version easily in one go
-project(prmon VERSION 1.1.1)
+project(prmon VERSION 1.2.0)
```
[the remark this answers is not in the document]
graeme-a-stewart: Let's do a minor one for now... I'd still like to see the open issues resolved before we are allowed to have 2.0.0 😄

package/CMakeLists.txt
```diff
@@ -3,12 +3,14 @@ find_package(nlohmann_json REQUIRED)
...
-           src/cpumon.cpp
-           src/countmon.cpp
-           src/wallmon.cpp
-           src/memmon.cpp)
+					 src/netmon.cpp
+					 src/iomon.cpp
+					 src/cpumon.cpp
+					 src/countmon.cpp
+					 src/wallmon.cpp
+					 src/memmon.cpp
```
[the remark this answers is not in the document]
graeme-a-stewart: Yes, some tabs got snuck in - converting them back to spaces

package/src/nvidiamon.cpp
```diff
@@ -0,0 +1,207 @@
...
+  const std::vector<std::string> cmd = {"nvidia-smi", "pmon", "-s",
+                                        "um",         "-c",   "1"};
+
+  for (auto& stat : nvidia_stats) stat.second = 0;
+
+  auto cmd_result = prmon::cmd_pipe_output(cmd);
+  if (cmd_result.first) {
+    // Failed
+    return;
+  }
```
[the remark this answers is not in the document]
graeme-a-stewart: Yes, for this one we shouldn't just ignore it. It means something really bad happened.

package/src/nvidiamon.cpp
graeme-a-stewart: I have no idea what that was supposed to be!

package/src/nvidiamon.cpp
graeme-a-stewart: Do you think it's really worth it for a private member variable?

package/src/nvidiamon.cpp
graeme-a-stewart: Do you think it's really worth it for a private member variable?

package/src/nvidiamon.cpp
graeme-a-stewart: Absolutely correct - when I was making this simpler I flipped the logic (or rather *failed* to flip it).

(I couldn't run any test jobs yesterday as the GPUs were busy, so missed this point!)

package/src/nvidiamon.cpp
graeme-a-stewart: Yes, this is correct

graeme-a-stewart: In fact, better to setup a constant that describes what we are doing than use a magic number

package/src/nvidiamon.cpp
amete: For some reason I thought they were public, which was another question, but indeed they are private and not exposed outside in any way. I agree, this is fine.

PR #129, opened 2020-06-12
Code formatting and cleanup; documentation

### LINKED ISSUE #127: Contribution and coding guide
We should add a contribution guide to the project encouraging other people. I think two things in particular we need

- A description of how to add a new monitoring component
- A code style guide

For the latter I looked at `clang-format` and all of the standard styles look pretty decent to me. I think we used the Google setup in the past, though probably our last PRs deviated from that a bit.

We could also add a CI job to check code formatting.

graeme-a-stewart:
Closes #127 

# Clang-format
Reformat all the code according to the Google style of `clang-format`.

Add also a new target to Cmake to run clang-format on all sources (`make clang-format`).
- I found an issue between different versions of clang-format, specifically v10 has options that v8 does not recognise, so it will not run
  - That meant that adding a v10 `.clang-format` file broke v8
    - v10 is what I get with brew on OS X; v8 is what's installed on CentOS8 by yum
- Thus the formatter is run with `--style=Google`

Hopefully this will not cause too much confusion (running v8 *after* v10 there are no changes to the files, so fingers crossed...)

# Code reorganisaion
I extracted a long-standing reorganisation of utilities that have been sitting in with the NVIDIA developments in the `nvidia-mon` branch (and make rebasing continually painful)
- Rename pidutils -> prmonutils (these became more general)
- Remove `SIGCHLD` handler in favour of reaping
- Remove useless `prmon.h` header
- Add `cmd_pipe_output()` utility

Some useless pieces of code were removed from `docs` (as we don't use doxygen).

# Documentation
- Add a code of conduct (lifted straight from <https://www.contributor-covenant.org>).
- Add a contribution guide
- Add some documentation on how monitors should be implemented

# Copyright and License
- Made sure that copyright statements exist and are up to date (2018-2020)
- Added a single line re. the license to each source file


graeme-a-stewart:
I think this is GTG. With this and the GitHub Actions in place I can rebase the GPU branch and then that should be able to (finally) go in.

package/src/prmon.cpp
```diff
@@ -12,71 +13,46 @@
...
-        std::clog << "Child process " << pid << " was stopped by signal"
-                  << WSTOPSIG(status) << std::endl;
-      else if (WIFCONTINUED(status))
-        std::clog << "Child process " << pid << " was continued" << std::endl;
-    }
-  }
-}
+bool prmon::sigusr1 = false;
 
 int MemoryMonitor(const pid_t mpid, const std::string filename,
```
amete, line 31: Maybe this might be a good time to finally convert this to `ProcessMonitor` instead of `MemoryMonitor`?

package/src/registry.h
```diff
@@ -1,90 +1,102 @@
-// Copyright (C) CERN, 2020
+// Copyright (C) 2018-2020 CERN
+// License Apache2 - see LICENCE file
+
+// Implementation of a class registry that is used
+// for all prmon monitoring classes
+//
+// Grateful acknowledgement to https://github.com/psalvaggio/cppregpattern
+// which inspired this implementation of the pattern
```
amete, line 8: I think this is great. Maybe we can also add an `doc/ACKNOWLEDGEMENTS.md` and collate the information there instead, which would be easier to list and identify.

package/src/prmon.cpp
graeme-a-stewart: Yes - ideal time to do that indeed. Done in last commit.

package/src/registry.h
graeme-a-stewart: Also a good idea - I rearranged a little between `ACKNOWLEDGEMENTS.md` and `AUTHORS.md` as well

I did keep it in the root directory as it's not really documentation, right?

graeme-a-stewart:
Oh flip, I didn't really mean to merge in from `master` - I kind of misread that button text. Should be harmless...

amete:
All looks good, merged 😄

PR #130, opened 2020-06-12
Putting in a basic GitHub Actions workflow

amete:
This is for playing around w/ GitHub actions, not to be merged just yet...


amete:
I think this is ready to be reviewed as a first implementation. With this PR we're dropping Travis completely and running all our CI jobs in GitHub actions. CodeFactor also runs as part of the checks. 

The real nice thing about GitHub actions is that it allows writing custom actions (i.e. see [here](https://help.github.com/en/actions/creating-actions)). As the next step, I think it would be good to write an action where we run `clang-format` and perhaps `clang-tidy` to automatically apply formatting/style corrections as part of the workflow. Since #129 introduces `clang-format` target for `cmake` configuration, perhaps this should be developed on top of that. I'll be looking into it in another PR soon.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+    steps:
+    # Checks-out our repository under $GITHUB_WORKSPACE, so our job can access it
+    - uses: actions/checkout@v2
+
+    # Pulls the associated Docker image
+    - name: Docker pull
+      run: |
+        if [ "$PLATFORM" == "c7-dev-py2" ];
+        then
+          docker pull hepsoftwarefoundation/c7-dev
```
graeme-a-stewart, line 38: Could it be made a little better here if we defined `DIMAGE` as the docker image to pull? And for `c7-dev-py2` also defined at this point `CMAKE_EXTRA=-DPYTHON_TEST=python2`? It seems a bit of a cleaner separation between PLATFORM and Docker image.

```
if [ "$PLATFORM" == "c7-dev_py2" ];
then
  DIMAGE=hepsoftwarefoundation/c7-dev
  CMAKE_EXTRA=-DPYTHON_TEST=python2
else
  DIMAGE=hepsoftwarefoundation/$PLATFORM
fi
docker pull $DIMAGE
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+        then
+          docker run -v $(pwd):/mnt hepsoftwarefoundation/$PLATFORM /mnt/.github/scripts/build-test.sh;
+        elif [ "$PLATFORM" == "u18-dev" ];
+        then
+          docker run -e CMAKE_EXTRA=-Dnlohmann_json_DIR=/usr/lib/cmake -v $(pwd):/mnt hepsoftwarefoundation/$PLATFORM /mnt/.github/scripts/build-test.sh;
+        elif [ "$PLATFORM" == "u20-dev" ];
+        then
+          docker run -e CXX=g++-9 -e CC=gcc-9 -v $(pwd):/mnt hepsoftwarefoundation/$PLATFORM /mnt/.github/scripts/build-test.sh;
+        else
+          echo "Uknown DIMAGE $DIMAGE"
```
graeme-a-stewart, line 62: `echo "Unknown Docker Image: $DIMAGE`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+        else
+          docker pull hepsoftwarefoundation/$PLATFORM
+        fi
+
+    # Builds the code and runs the test
+    - name: Build and Test
+      run: |
+        if [ "$PLATFORM" == "c7-dev-py2" ];
+        then
+          docker run -e CMAKE=cmake3 -e CMAKE_EXTRA=-DPYTHON_TEST=python2 -v $(pwd):/mnt hepsoftwarefoundation/c7-dev scl enable devtoolset-8 /mnt/.github/scripts/build-test.sh;
```
graeme-a-stewart, line 48: Update to use $DIMAGE, if you agree with the above suggestion

graeme-a-stewart:
> The real nice thing about GitHub actions is that it allows writing custom actions (i.e. see [here](https://help.github.com/en/actions/creating-actions)). As the next step, I think it would be good to write an action where we run `clang-format` and perhaps `clang-tidy` to automatically apply formatting/style corrections as part of the workflow. Since #129 introduces `clang-format` target for `cmake` configuration, perhaps this should be developed on top of that. I'll be looking into it in another PR soon.

Sounds good to me - and it gives a positive reason to move to GitHub actions over Travis.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+jobs:
+  # Main building and testing
+  build-test:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Define the platforms to run on. Later determines the Docker image that'll be used
+    strategy:
+      matrix:
+        platform: [ c7-dev-py2, c7-dev, c8-dev, u18-dev, u20-dev ]
```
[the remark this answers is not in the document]
amete: They seem to use this `strategy/matrix` syntax to create a unique job for each combination. This was the easiest way to set things up from what I can see. In the latest update, I changed this bit to `c7-dev_py2` (to make it a bit clearer that `py2` is an option) and later on define `DIMAGE` and `CMAKE_EXTRA` based on the value of this guy.

.github/workflows/ci.yml
amete: This logic is updated now. Setting environment variables in different steps is a bit weird if you want to carry it over to other steps. But, in essence we do exactly this now.

.github/workflows/ci.yml
amete: Done.

.github/workflows/ci.yml
amete: Done.

amete:
Thanks a lot Graeme. Good suggestions, as always. I adopted all and update the necessary bits. Please let me know how it looks now, many thanks 😄

.github/workflows/ci.yml
graeme-a-stewart: One thing that grates for me still is the way we set specific environment variable in the `docker run` command, but I can't quite see an easy way out of that right now 🤔 

I think this, at least, means that in one place it's clear what `_py2` means, which is a good thing.

graeme-a-stewart:
All good now - thanks a lot!

drbenmorgan:
Sorry, know this is closed, but only just saw the mail - just to advertise (via the UK RSE Slack), that there's an exemplar action for clang-tidy here:

https://github.com/ZedThree/clang-tidy-review

might be useful or provide some inspiration,

amete:
> Sorry, know this is closed, but only just saw the mail - just to advertise (via the UK RSE Slack), that there's an exemplar action for clang-tidy here:
> 
> https://github.com/ZedThree/clang-tidy-review
> 
> might be useful or provide some inspiration,

Thanks a lot Ben. We'll increase coverage of the checks, possibly including `clang-format` and `clang-tidy` workflows. This'll be very useful indeed.

PR #131, opened 2020-06-16
Adding a basic clang-format GitHub action

amete:
For running some tests, not to be merged yet...


amete:
OK, the first prototype seems to be working. Here we rely on a custom GitHub action that I wrote: [amete/clang-format-action](https://github.com/amete/clang-format-action). 

The action is fairly specific for our project, where I take advantage of the new `clang-format` target that is introduced #129. Right now what we do is:

- For each push, we run `clang-format` for the entire repository
- Add a new commit w/ the changes (if any) as you can see in [7e54689]
(7e54689)

I'm not super sure why the CI jobs get stuck after the automatic push though. That needs to be understood.

amete:
Hi @graeme-a-stewart,

I think this is ready to be reviewed. I changed the workflow a bit such that `clang-format` will now run only when there is a push into either `master` or `stable` branches. For example, when a given PR is accepted into `master`, `clang-format` will run in a custom docker workflow as explained above, and commit the changes (if any) on top of the latest commit w/ `clang-format` as the author and `Apply clang-format` as the commit message. 

What do you think?

graeme-a-stewart:
Hi @amete 

Thanks a lot for this - it looks really nice. I have a few questions, possibly related to my ignorance about GitHub actions.

- Does this build the container every time from `ubuntu:latest`? That would seem a bit wasteful. 
- I'm a bit worried about `ubuntu:latest`changing under our feet, would it not be safer to run this in a known version (`ubuntu:focal` would be what I would go for today, i.e. 20.04)
- I also saw that you build and install `nlohmann/json`, but this is available in Ubuntu 20 as `nlohmann-json3-dev`
- It's not very clear to me what happens if there are no changes - it looks in `entrypoint.sh` like you do a commit and a push anyway
  - Could we instead detect no changes, print a comforting message and exit?

amete:
Hi @graeme-a-stewart,

> Hi @amete
> 
> Thanks a lot for this - it looks really nice. I have a few questions, possibly related to my ignorance about GitHub actions.

I'm also learning this as I go 😄 

> * Does this build the container every time from `ubuntu:latest`? That would seem a bit wasteful.

Indeed this seems to be their workflow. Essentially what happens is that they build the image from the `Dockerfile` (every time), the execute `entrypoiny.sh`. Then in the workflow, one can further define various things that should be run, similar to what we do in `ci.yml`. They don't seem to be bothered by this. If we end up running only when there is a push to `master`/`stable`, then this might be OK, I believe.

Another option might be how we do the build. In other words, we can create our own container, then in the workflow we simply run it and execute the script. That'll be faster and wouldn't require defining an action ourselves. Now thinking about it, maybe that's a better way to go forward? Does our `u20-dev` already include `clang-format`?

> * I'm a bit worried about `ubuntu:latest`changing under our feet, would it not be safer to run this in a known version (`ubuntu:focal` would be what I would go for today, i.e. 20.04)

That's a good point indeed. I can change this.

> * I also saw that you build and install `nlohmann/json`, but this is available in Ubuntu 20 as `nlohmann-json3-dev`

Oh, didn't realize this, I'll drop building it myself and simply install `nlohmann-json3-dev`

> * It's not very clear to me what happens if there are no changes - it looks in `entrypoint.sh` like you do a commit and a push anyway
>   
>   * Could we instead detect no changes, print a comforting message and exit?

I can most certainly play w/ this part. But in essence you're correct. I kind of piggybacked on the fact that `git` is smart enough to know that when there is no change whatsoever, there is nothing to commit/push so it simply says `nothing to commit, working tree clean` and `Everything up-to-date`. We can stick in a `git diff` etc. to be a bit more verbose though.

amete:
Himm, looks like running the `make clang-format` from within the `u20-dev` container core dumped, which is something I cannot reproduce locally:

https://github.com/HSF/prmon/pull/131/checks?check_run_id=777902347

I'll have to check this again tomorrow, but this is essentially my other suggestion @graeme-a-stewart.

amete:
OK, I think the problem is that we cannot write into the source directory from within our container (which is what `make clang-format` tries to do).  

On the test machine we simply do:

```
docker run -v $(pwd):/mnt hepsoftwarefoundation/u20-dev /mnt/.github/scripts/clang-format.sh
```

then `make clang-format` tries to overwrite files under `/mnt` which is not allowed. When I run the above command I can actually overwrite the files. Maybe they don't allow this due to security concerns.

I'm starting to think going the custom actions way might be a better option after all.

amete:
Hi @graeme-a-stewart, 

As it's obvious from above, after a bit of a battle w/ the system I think I have an improved version. Running the formatting inside a container in the workflow doesn't seem to work because the source code is read-only by the looks of it. Therefore, I went back to using the action. However, I have a new tag for it: [amete/clang-format-action/v0.2](https://github.com/amete/clang-format-action/tree/v0.2):

-  Now we use `ubuntu:focal` for better control
-  I completely dropped the `cmake` approach and made the action as generic as possible so that we can actually use it in other projects as well
-  Now we don't try to commit/push unless there is a change
-  On the workflow side, we can now configure the style (defaults to Google) and retrieve and print a message as to what happens, e.g. see [here](https://github.com/HSF/prmon/runs/779759528?check_suite_focus=true)

I configured it such that the action will run whenever we push into `master`/`stable`. This way we don't interfere w/ the PRs.

I think this is a cleaner and better approach overall. Once we're happy we can move the action to somewhere under HSF.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches: [ master, stable ]
+
+# Defines the main job where we checkout our code, run clang-format
+# and commit changes
+jobs:
+  clang-format:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: amete/clang-format-action@v0.1
```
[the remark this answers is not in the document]
amete: This I completely agree, please see below 👍

graeme-a-stewart:
@amete - looks good to go for me as well. I approve of all your changes!

Shall I fork your actions repo to HSF, then we make a last change to take the action from there?

amete:
> @amete - looks good to go for me as well. I approve of all your changes!
> 
> Shall I fork your actions repo to HSF, then we make a last change to take the action from there?

Thanks a lot @graeme-a-stewart. Yep, that sounds good to me. 

Once this goes in, I plan to create a dummy PR w/ a synthetic formatting change and we can test it in full action.

graeme-a-stewart:
Repo forked to:

https://github.com/HSF/clang-format-action

@amete you have admin rights too

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,28 @@
...
+  clang-format:
+    runs-on: ubuntu-latest
+    steps:
+    # Checks-out our repository under $GITHUB_WORKSPACE, so our job can access it
+    - uses: actions/checkout@v2
+
+    # Applies clang-format to all C/C++ files w/ the specified style
+    - name: Apply clang-format
+      id: formatting
+      uses: amete/clang-format-action@v0.2
```
graeme-a-stewart, line 22: Update to `HSF/clang-format-action@v0.2`

amete: Done.

amete:
Thanks a lot @graeme-a-stewart. Let's see how this'll work.

PR #134, opened 2020-06-17
Omit running clang-format for pushes into master/stable

amete:
With #133 it became obvious that the logic that we discussed in #131 doesn't work because `master` branch is protected according to our settings. Loosening the requirements will resolve this but I don't think we want to do that just for this.

This PR is proposing to change how we run `clang-format`. Now, we run it for all pushes except for the `master` and `stable` branches. So, when a new branch is pushed to the repository, `clang-format` will run automatically and amend the branch if necessary, as it just happened at:

https://github.com/HSF/prmon/runs/780586384?check_suite_focus=true

Then any consecutive push will trigger `clang-format` and fix issues as they appear. One has to be careful and `pull` the changes locally iff `clang-format` actually updated branch. I don't think this is the end of the world but let me know what you think @graeme-a-stewart.

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,11 +1,11 @@
 # Name of the workflow
 name: Clang-Format
 
-# Controls when the action will run. Triggers the workflow on push for
-# the master and stable branches
+# Controls when the action will run. Triggers the workflow on push
+# except for the master and stable branches
 on:
   push:
-    branches: [ master, stable ]
+    branches-ignore: [ master, stable ]
 
 # Defines the main job where we checkout our code, run clang-format
 # and commit changes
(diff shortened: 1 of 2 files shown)


amete:
As one can see the formatting issues that is synthetically introduced in #133 is now resolved w/ d07ef10.

graeme-a-stewart:
Yep, I agree that relaxing the restrictions on `master` would be unwise. This method allows us to also see the effects of clang-format before merging and that's probably a good thing!

We usually squash when applying to master, so then having a few of these technical commits in the branch history is not a problem, so the whole workflow seems sound to me.

If developers have clang-formatted in advance of their push then they won't need to pull, which is an incentive to get it correct on their side.

Issue #135, opened 2020-06-18
prmon broken on ARM


graeme-a-stewart:
I tried compiling prmon on ARM (Raspberry Pi 4 with Raspberry Pi OS). Perhaps unsurprisingly there are a few issues to fix:

- Casting the return value of `getopt_long()` to `char` we implicitly assume `signed char` (comparing to `-1`), but actually plain `char` is system dependent and on ARM it's unsigned, meaning that `char(-1)` is `255`
- The CPU hardware counters come out with `0` core count
  - Specifically we need to fix this for ARM, plus the code needs to be protected against division by zero
- The compiler on ARM spits out thousands of ABI warnings (which are irrelevant), so on this platform we need to add the `-Wno-psabi`

amete:
I guess the second can be easily avoided with:

`
 unsigned int nSockets = ( nSiblings != 0 ? nCPU / nSiblings : 0 );
 unsigned int nThreads = ( nCores != 0 ? nSiblings / nCores : 0 );
`

at https://github.com/HSF/prmon/blob/master/package/src/cpumon.cpp#L108-L109. I guess I was not expecting the number of cores to be zero 😃 

Just out of curiosity, what does the `/proc/cpuinfo` file look like? If it doesn't exist then we shouldn't have made that far.

graeme-a-stewart:
It does exist, but it's rather different from what we get on x86:

[cpuinfo.txt](https://github.com/HSF/prmon/files/4797284/cpuinfo.txt)

amete:
That's an interesting looking `/proc/cpuinfo`, never seen such a format before but I'm very biased 😄 

I guess one thing we can do is to check for "ARM" in the model name and try not to fill `nSockets`, `nCoresPerSocket `, and `nThreadsPerCore` if it exists. There doesn't seem to be any information that would/could replace this information. For ARM, we'll have only the model name along w/ the `nCPU`. 

Do you want me to take a stab or are you already on it? I'm happy either way 😄

graeme-a-stewart:
> I guess the second can be easily avoided with:
> 
> `unsigned int nSockets = ( nSiblings != 0 ? nCPU / nSiblings : 0 );`
> `unsigned int nThreads = ( nCores != 0 ? nSiblings / nCores : 0 );`
> 
> [here](https://github.com/HSF/prmon/blob/master/package/src/cpumon.cpp#L108-L109). I guess I was not expecting the number of cores to be zero 😃

[...]

PR #139, opened 2020-06-18
Get the CPU information from lscpu

### LINKED ISSUE #136: Update to retrieve CPU information using lscpu
As noted in #135 it will be more robust to get CPU/socket/thread/core counts from `lscpu` as it is tested on many more architectures than we have available (use `prmon::cmd_pipe_output` to run the external command). The current implementation crashes on ARM machines.

amete:
As we discussed in #135 using `lscpu` instead of manually reading/parsing `/proc/cpuinfo` is more robust against various architectures (especially non-x86). This is especially easy now given that we have `prmon::cmd_pipe_output` to invoke any command. On my laptop, the output looks like this now:

```
{
  ...  
  "HW": {
    "cpu": {
      "Core(s) per socket": 1,
      "Model name": "Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz",
      "Socket(s)": 4,
      "Thread(s) per core": 1
    },  
    "mem": {
      "MemTotal": 2038904
    }   
  },  
  ...
}
```

Closes #136.

### DIFF
--- a/package/src/cpumon.cpp
+++ b/package/src/cpumon.cpp
@@ -112,9 +112,9 @@ void const cpumon::get_hardware_info(nlohmann::json& hw_json) {
     // Fill the JSON with the information
     if (metricToName.count(key) == 1) {
       if (isNumber(value))
-        hw_json["HW"]["cpu"][key] = std::stoi(value);
+        hw_json["HW"]["cpu"][metricToName.at(key)] = std::stoi(value);
       else
-        hw_json["HW"]["cpu"][key] = value;
+        hw_json["HW"]["cpu"][metricToName.at(key)] = value;
     }
   }
 
--- a/package/src/cpumon.cpp
+++ b/package/src/cpumon.cpp
@@ -63,60 +63,56 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
 
 // Collect related hardware information
 void const cpumon::get_hardware_info(nlohmann::json& hw_json) {
-  // Read some information from /proc/cpuinfo
-  std::ifstream cpuInfoFile{"/proc/cpuinfo"};
-  if (!cpuInfoFile.is_open()) {
-    std::cerr << "Failed to open /proc/cpuinfo" << std::endl;
+  // Define the command and run it
+  const std::vector<std::string> cmd = {"lscpu"};
+
+  auto cmd_result = prmon::cmd_pipe_output(cmd);
+
+  // If the command failed print an error and move on
+  if (cmd_result.first) {
+    std::cerr << "Failed to execute 'lscpu' to get CPU information (code "
+              << cmd_result.first << ")" << std::endl;
     return;
   }
 
-  // Metrics to read from the input
-  std::vector<std::string> metrics{"processor", "model name", "siblings",
-                                   "cpu cores"};
-  unsigned int nCPU = 0, nSiblings = 0, nCores = 0;
+  // Metrics that'll be kept from the lscpu output 
+  const std::vector<std::string> metrics{"Model name",  "Socket(s)", 
+    "Core(s) per socket", "Thread(s) per core"};
 
-  // Loop over the file
-  std::string line;
-  while (std::getline(cpuInfoFile, line)) {
+  // Useful function to determine if a string is purely a number
+  auto isNumber = [] (const std::string& s) 
+  {
+    return !s.empty() && std::find_if(s.begin(), 
+          s.end(), [](unsigned char c) { return !std::isdigit(c); }) == s.end();
+  };
+
+  // Loop over the output, parse the line, check the key and store the value if
+  // requested
+  std::string key{}, value{};
+
+  for (const auto& line : cmd_result.second) {
+
+    // Continue on empty line
     if (line.empty()) continue;
+
+    // Tokenize by ":"
     size_t splitIdx = line.find(":");
-    std::string val;
-    if (splitIdx != std::string::npos) {
-      val = line.substr(splitIdx + 1);
-      if (val.empty()) continue;
-      for (const auto& metric : metrics) {
-        if (line.size() >= metric.size() &&
-            line.compare(0, metric.size(), metric) == 0) {
-          if (metric == "processor")
-            nCPU++;
-          else if (metric == "siblings") {
-            if (nSiblings == 0) nSiblings = std::stoi(val);
-          } else if (metric == "cpu cores") {
-            if (nCores == 0) nCores = std::stoi(val);
-          } else if (hw_json["HW"]["cpu"][metric].empty()) {
-            hw_json["HW"]["cpu"][metric] =
-                std::regex_replace(val, std::regex("^\\s+|\\s+$"), "");
-          }
-        }  // end of metric check
-      }    // end of populating metrics
-    }      // end of seperator check
-  }        // end of reading cpuInfoFile
+    if (splitIdx == std::string::npos) continue;
 
-  // The clear assumption here is that there is a single type of processor
-  // installed nCPU = nSockets * nSiblings nSiblings = nCoresPerSocket *
-  // nThreadsPerCore
-  // Next 4 lines are a workaround for Raspberry Pi
-  if (!nSiblings) nSiblings = nCPU;
-  if (!nCores) nCores = nCPU;
-  unsigned int nSockets = nCPU / nSiblings;
-  unsigned int nThreads = nSiblings / nCores;
-  hw_json["HW"]["cpu"]["nCPU"] = nCPU;
-  hw_json["HW"]["cpu"]["nSockets"] = nSockets;
-  hw_json["HW"]["cpu"]["nCoresPerSocket"] = nCores;
-  hw_json["HW"]["cpu"]["nThreadsPerCore"] = nThreads;
+    // Read "key":"value" pairs
+    key = line.substr(0, splitIdx);
+    value = line.substr(splitIdx+1);
+    if ( key.empty() || value.empty() ) continue;
+    key = std::regex_replace(key, std::regex("^\\s+|\\s+$"), "");
+    value = std::regex_replace(value, std::regex("^\\s+|\\s+$"), "");
 
-  // Close the file
-  cpuInfoFile.close();
+    // Fill the JSON with the information
+    for (const auto& metric : metrics) {
+      if ( key.compare(metric) != 0 ) continue;
+      if ( isNumber(value) ) hw_json["HW"]["cpu"][key] = std::stoi(value);
+      else hw_json["HW"]["cpu"][key] = value;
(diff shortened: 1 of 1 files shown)


package/src/cpumon.cpp
```diff
@@ -63,60 +63,57 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
...
+              << cmd_result.first << ")" << std::endl;
     return;
   }
 
-  // Metrics to read from the input
-  std::vector<std::string> metrics{"processor", "model name", "siblings",
-                                   "cpu cores"};
-  unsigned int nCPU = 0, nSiblings = 0, nCores = 0;
+  // Metrics that'll be kept from the lscpu output
+  const std::vector<std::string> metrics{
```
graeme-a-stewart, line 79: We should add the total number of CPUs as well (`CPU(s)`)

package/src/cpumon.cpp
```diff
@@ -63,60 +63,57 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
...
+
+    // Read "key":"value" pairs
+    key = line.substr(0, splitIdx);
+    value = line.substr(splitIdx + 1);
+    if (key.empty() || value.empty()) continue;
+    key = std::regex_replace(key, std::regex("^\\s+|\\s+$"), "");
+    value = std::regex_replace(value, std::regex("^\\s+|\\s+$"), "");
+
+    // Fill the JSON with the information
+    for (const auto& metric : metrics) {
```
graeme-a-stewart, line 109: This is quadratic in complexity, but given that the output is small and the metrics are small in number I think we can get away with it (also, it's only called once)

graeme-a-stewart:
@amete, thanks a lot.

The only additional comment I have is that the key names produced by `lscpu` are quite ugly - they have spaces and parentheses in them. I would be inclined to simplify them to:

```
CPUs
Sockets
CoresPerSocket
ThreadsPerCore
```

That makes them easier for a client to handle. I am aware that this is a small extra level of complexity (I'd suggest a `map(lscpu key -> our key)`), but I think it's nicer.

What do you think?

amete:
I agree @graeme-a-stewart, we can easily map the names as you suggested.

graeme-a-stewart:
Also, on my Pi4 I get:

```
    "cpu": {
      "Core(s) per socket": 4,
      "Model name": "Cortex-A72",
      "Socket(s)": 1,
      "Thread(s) per core": 1
    },
```

Which is exactly what we wanted.

package/src/cpumon.cpp
amete: That was my line of thinking as well. However, given that we have only a handful of variables (conditions), I can just as well put in an if statement here. What do you think?

package/src/cpumon.cpp
amete: Done.

package/src/cpumon.cpp
amete: I switched to using a simple `if` here.

amete:
I just pushed a new version @graeme-a-stewart, let me know what you think, many thanks 😄

amete:
BTW, this is how this block looks now:

```
   "cpu": {
      "CPUs": 4,
      "CoresPerSocket": 1,
      "ModelName": "Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz",
      "Sockets": 4,
      "ThreadsPerCore": 1
    },
```

amete:
Another BTW, running within an Ubuntu based container on my MacBook Pro seems to have incorrect `/proc/cpuinfo` about the hardware. I have a single quad-core Intel i5 but `/proc/cpuinfo` publishes this as four single-core processors. That's interesting.

graeme-a-stewart:
> BTW, this is how this block looks now:
> 
> ```
>    "cpu": {
>       "CPUs": 4,
>       "CoresPerSocket": 1,
>       "ModelName": "Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz",
>       "Sockets": 4,
>       "ThreadsPerCore": 1
>     },
> ```

Wait... that can't be correct. How can you have a 4 socket machine in your laptop?

Doing a few checks on my platforms

graeme-a-stewart:
OK... I think it's actually an issue with Docker on OS X. The VM layer is translating each CPU that is given to Docker as a separate socket. For both bare metal and Docker on Linux I get what I expect:

```
    "cpu": {
      "CPUs": 16,
      "CoresPerSocket": 8,
      "ModelName": "Intel(R) Core(TM) i7-7820X CPU @ 3.60GHz",
      "Sockets": 1,
      "ThreadsPerCore": 2
    },
```

And it's fine on the Pi too:

```
    "cpu": {
      "CPUs": 4,
      "CoresPerSocket": 4,
      "ModelName": "Cortex-A72",
      "Sockets": 1,
      "ThreadsPerCore": 1
    },
```

In fact the older version of the parser, directly from `/proc/cpuinfo` also reports the same when run on OS X Docker.

So it's not at all a problem we need to worry about in practice.

package/src/cpumon.cpp
```diff
@@ -106,12 +110,12 @@ void const cpumon::get_hardware_info(nlohmann::json& hw_json) {
     value = std::regex_replace(value, std::regex("^\\s+|\\s+$"), "");
 
     // Fill the JSON with the information
-    for (const auto& metric : metrics) {
-      if (key.compare(metric) != 0) continue;
+    if (key == "Model name" || key == "CPU(s)" || key == "Socket(s)" ||
```
graeme-a-stewart, line 113: Oh no... I actually much preferred the previous version. Now you have these key names encoded twice, *in different places*, which is bad.

I actually doubt that it's any more efficient than the previous `for`loop (which can be slightly improved by adding a `break` after you process the match, because each key can only match once).

amete: No problem, let me know how it looks now 👍

package/src/cpumon.cpp
```diff
@@ -63,60 +63,62 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
...
+
+    // Read "key":"value" pairs
+    key = line.substr(0, splitIdx);
+    value = line.substr(splitIdx + 1);
+    if (key.empty() || value.empty()) continue;
+    key = std::regex_replace(key, std::regex("^\\s+|\\s+$"), "");
+    value = std::regex_replace(value, std::regex("^\\s+|\\s+$"), "");
+
+    // Fill the JSON with the information
+    for (const auto& metric : metricToName) {
```
graeme-a-stewart, line 113: Oh but wait... now we can use the map:

```cpp
if (metricToName.count(metric) == 1) {
...
}
```

That's the right way to do it!

amete: Good point, changed now. I seldom use maps, as a deliberate choice, so this completely escaped me. Is this still an `O(N)` operation or is it actually smarter than that?

amete: Wait I got it wrong 😄

amete: OK, we should be good now, today is not my day I guess.

package/src/cpumon.cpp
```diff
@@ -63,60 +63,60 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
...
-  // Metrics to read from the input
-  std::vector<std::string> metrics{"processor", "model name", "siblings",
-                                   "cpu cores"};
-  unsigned int nCPU = 0, nSiblings = 0, nCores = 0;
-
-  // Loop over the file
-  std::string line;
-  while (std::getline(cpuInfoFile, line)) {
+  // Map lscpu names to the desired ones in the JSON
+  const std::map<std::string, std::string> metricToName{
```
graeme-a-stewart, line 79: Although it's not a big deal, make this an `unordered_map`... (just good practice)

package/src/cpumon.cpp
```diff
@@ -63,60 +63,60 @@ std::map<std::string, double> const cpumon::get_json_average_stats(
...
+    key = line.substr(0, splitIdx);
+    value = line.substr(splitIdx + 1);
+    if (key.empty() || value.empty()) continue;
+    key = std::regex_replace(key, std::regex("^\\s+|\\s+$"), "");
+    value = std::regex_replace(value, std::regex("^\\s+|\\s+$"), "");
+
+    // Fill the JSON with the information
+    if (metricToName.count(key) == 1) {
+      if (isNumber(value))
+        hw_json["HW"]["cpu"][metricToName.at(key)] = std::stoi(value);
```
graeme-a-stewart, line 115: Isn't `metricToName[key]` better? Although `at` gives a bounds check, we know already that the key is present. And if it weren't then it would throw `std::out_of_range` that we don't catch and the program would crash anyway, so it's no protection.

package/src/cpumon.cpp
graeme-a-stewart: I think we are both ready for the weekend!

graeme-a-stewart: > Good point, changed now. I seldom use maps, as a deliberate choice, so this completely escaped me. Is this still an `O(N)` operation or is it actually smarter than that?

As it's a map (with a hash) it's usually constant (with the worst case being linear).

graeme-a-stewart: The main advantage here of using a map is it makes the program's intention clearer than looping over a vector, which means less chance of screwing up... 😁

package/src/cpumon.cpp
amete: operator `[]` is not declared as const for `std::map` (or `std::unordered_map` for that matter), so that gives me a compilation error since the method is `cont`. That's why I ended up using `at`:

```
/home/amete/github/prmon/package/src/cpumon.cpp: In member function 'virtual const void cpumon::get_hardware_info(nlohmann::json&)':
/home/amete/github/prmon/package/src/cpumon.cpp:117:46: error: passing 'const std::unordered_map<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >' as 'this' argument discards qualifiers [-fpermissive]
  117 |         hw_json["HW"]["cpu"][metricToName[key]] = value;
```

Let me know if I'm missing something here.

graeme-a-stewart: You are right - I missed that! `[]` is also used a setter so it's not `const`. Yes, just leave it as is. Sorry for the noise!

package/src/cpumon.cpp
amete: Done.

package/src/cpumon.cpp
amete: We re-wrote this bit, so I'm resolving.

amete:
Let me know how it all looks now @graeme-a-stewart, many thanks 👍

graeme-a-stewart:
Thanks a lot! All LGTM 👍

PR #144, opened 2020-07-01
Implement units information

### LINKED ISSUE #126: Describe units
We now have a large amount of monitoring in prmon and, obviously this is data that we give as *numbers*. However, we don't describe properly what the *units* are, and this is bad.

This should be fixed.

An idea would be that in `utils.h`, where we setup the column names for monitoring, we'd also record the units. Then we can have an option to *describe* the units, either as text (for humans) or as a JSON file (for machines). Note we need units for the snapshot/max values and for the average values. They may not be the same.

graeme-a-stewart:
Added a new helper class `parameter` that stores a parameter and its associated units. Each concrete monitor now defines its monitored parameters as a vector of these classes (moving them out of `utils.h` as they really were private to each class). This avoids adding new parameters and forgetting units.

Note that for units which are pure numbers the unit describes what the number represents (e.g., packets or threads).

If a unit is an empty string it's suppressed as it corresponds to a metric which is not given (this happens a few times where the average value makes no sense).

To extract units a new function is added to the interface class, `get_unit_info`, that functions like the hardware information, taking the JSON object as a parameter and modifying it. As this is basically the same for all *concrete* monitors, this is in practice implemented via the `prmon::fill_units` helper function.

Printing units is not enabled by default, but only when the `--units/-u` option is specified.

The `testCPU.py` test function is enhanced to be able to check that the units output correspond correctly to the units in the JSON summary file and a new unit test for this is added.

A small enhancement to the `clang-format` setup now allows the binary to be specified to CMake with `-DCLANG_FORMAT=...`, which is useful on platforms that can have multiple versions (e.g. `clang-format-9` in Raspberry Pi OS cf. `clang-format` which is v7).

Closes #126


package/src/utils.cpp
```diff
@@ -95,4 +94,13 @@ const std::pair<int, std::vector<std::string>> cmd_pipe_output(
   return ret;
 }
 
-}  // namespace prmon
+const void prmon::fill_units(nlohmann::json& unit_json,
+                             const parameter_list& params) {
+  for (const auto& param : params) {
+    if (!((param.get_max_unit()).empty()))
```
amete, line 100: I think we have a redundant set of parentheses here, not that it really matters :)

package/src/wallmon.h
```diff
@@ -16,10 +16,16 @@
 #include <vector>
 
 #include "Imonitor.h"
+#include "parameter.h"
 #include "registry.h"
 
 class wallmon final : public Imonitor {
  private:
+  // const static std::vector<std::string> default_wall_params{"wtime"};
```
amete, line 24: I think we can delete this line.

package/src/countmon.h
```diff
@@ -13,10 +13,15 @@
...
 
 #include "Imonitor.h"
+#include "parameter.h"
 #include "registry.h"
 
 class countmon final : public Imonitor {
  private:
+  // Setup the parameters to monitor here
+  const prmon::parameter_list params = {{"nprocs", "n_procs", "n_procs"},
+                                        {"nthreads", "n_threads", "n_threads"}};
```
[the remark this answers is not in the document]
graeme-a-stewart: Yes, I was scratching my head to think what is best to have here. `1` would then be consistent with the plotting script.

PR #145, opened 2020-07-01
Relax test checks on numerical values

### LINKED ISSUE #138: Randomly failing tests
I noticed that our standard tests occasionally fail but succeed after a re-try. I don't think there is any obvious pattern (i.e. random tests fail at random times) and I'm sure I've seen this both locally and also in the CI. We need to take a closer look, maybe we don't run long enough or need to tweak the conditions. Have you also noticed this @graeme-a-stewart?

graeme-a-stewart:
As the tests run are quite short jitter makes them
fail occasionally (especially in Docker containers
on OS X for some reason). This change relaxes a bit
the numerics to cut these false positives.

Closes #138

### DIFF
--- a/package/tests/testCPU.py
+++ b/package/tests/testCPU.py
@@ -46,9 +46,9 @@ def setupConfigurableTest(threads=1, procs=1, child_fraction=1.0, time=10.0, sla
                 # CPU time tests
                 totCPU = prmonJSON["Max"]["utime"] + prmonJSON["Max"]["stime"]
                 expectCPU = (1.0 + (procs-1)*child_fraction) * time * threads
-                self.assertLess(totCPU, expectCPU, "Too high value for CPU time "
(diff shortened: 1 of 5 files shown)


graeme-a-stewart:
Hi @amete 

Regarding the IO test, the issue with `read_bytes` and `write_bytes` is that they are affected by caches, so in fact you can get 0 and it's not an error. If `rchar` and `wchar` are correct then `/proc` is being read properly.

Likewise for the net tests, the numbers of packets depend on how the network is configured on the machine (packet sizes for many grid nodes are set to jumbo, 8kB), so it's quite hard to guess what the correct answer is.

I also worry that we don't have a GPU test, but it will only be possible to buld the GPU burner if we have the CUDA SDK, so it's going to make things a bit more complex - but thanks a lot for creating #147 to discuss further.

PR #154, opened 2020-07-20
Bugfix in axis unit consistency check in the plotting script

amete:
I just realized a bug in the plotting script where we determine if the plotted variable and the unit that's set by the user is inconsistent. Essentially, in the current implementation these very reasonable inputs trigger the logic:

```
prmon_plot.py --input prmon.txt --xvar wtime --yvar utime,stime --yunit SEC --diff --stacked --otype pdf;
WARNING :: Changing yunit from SEC to SEC for consistency
INFO    :: Saved output into PrMon_wtime_vs_diff_utime_stime.pdf
```

and

```
prmon_plot.py --input prmon.txt --xvar wtime --yvar rchar,read_bytes,wchar,write_bytes --yunit MB --otype pdf;
WARNING :: Changing yunit from MB to B for consistency
INFO    :: Saved output into PrMon_wtime_vs_rchar_read_bytes_wchar_write_bytes.pdf
```

This PR should fix that by explicitly introducing a set of allowed units for each variable, and make a direct check of the input against that.

### DIFF
--- a/package/scripts/prmon_plot.py
+++ b/package/scripts/prmon_plot.py
@@ -18,32 +18,11 @@ except ImportError:
     sys.exit(-1)
 
 # Define the labels/units for beautification
-axisunits = {'vmem':'kb',
-             'pss':'kb',
-             'rss':'kb',
-             'swap':'kb',
-             'utime':'sec',
-             'stime':'sec',
-             'wtime':'sec',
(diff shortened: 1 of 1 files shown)


amete:
Thanks a lot @graeme-a-stewart :)

PR #161, opened 2020-07-28
Limit decimal places in JSON output of averages

### LINKED ISSUE #158: Round average values to more resonable number of DPs
When we changed to holding average values as floating point numbers (#108, #110) it was noted that the behaviour of `nlohmann::json` is to maintain as little loss of precision as possible in the output JSON number, often leading to many many useless decimal places of precision, e.g.,

```json
  "Avg": {
    "nprocs": 2.0,
    "nthreads": 6.0,
    "pss": 698.0,
    "rchar": 7568.472222222223,
    "read_bytes": 0.0,
    "rss": 4856.0,
    "rx_bytes": 4302.5,
    "rx_packets": 45.0,
    "swap": 0.0,
    "tx_bytes": 16323.75,
    "tx_packets": 35.69444444444444,
    "vmem": 40880.0,
    "wchar": 11.11111111111111,
    "write_bytes": 0.0
  },
```

This isn't very desirable, so we could truncate by doing a simple rounding operation, e.g.,

```cpp
json_summary["Avg"][(stat.first).c_str()] = long(stat.second*1000)/1000.;
```

That would do the rounding in a uniform way (the easiest to implement), but we could decide to do this on a per-metric basis or in a smart way where the absolute value is used to determine the rounding (e.g., some smart use of `std::modf`)

graeme-a-stewart:
As nlohmann::json always attempts to maintain full precision
when converting from a C++ float/double to a JSON number
it prints way too many decimal places in many cases.

Use std::modf to split the double precision value into integer
and fractional pieces and then truncate the DPs if needed.

Note that it's quite tricky to do this, so the fractional piece is
stored as a volatile to prevent the 'int(value*N)/double(N)' from
being optimised away. Note also that this is quite an expensive
operation, but in the context of prmon we can afford this at the
typical monitoring cadence.

Closes #158

### DIFF
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -131,16 +131,16 @@ int ProcessMonitor(const pid_t mpid, const std::string filename,
           auto wallclock_time = wallclock_monitor_p->get_wallclock_clock_t();
           for (const auto& stat :
                monitor.second->get_json_average_stats(wallclock_time)) {
-            // We will limit the accuracy here to per-mil as it doesn't
+            // We will limit the decimal place accuracy here as it doesn't
             // make any sense to provide tens of decimal places of
             // precision (this is a limitation of the nlohmann::json
             // library, which is always striving for minimal precision
             // loss when converting float types to decimal strings)
             double integer_piece;
-            // Volatile to prevent optimisation if rounding
+            // Volatile to prevent optimisation if performing rounding
             volatile double fractional_piece;
             fractional_piece = std::modf(stat.second, &integer_piece);
-            // The rounding method casting to int, then back to double is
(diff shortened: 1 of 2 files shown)


graeme-a-stewart:
Hi @amete - yes, you're right, it's printing a rounded integer for > 1000 and truncating to 3DP for <1000.

I wanted to avoid complexities like calculating `log10` to get more exact significant figure rounding.

If you also agree this is good enough I will update the description to be more accurate of the actual method.

amete:
> If you also agree this is good enough I will update the description to be more accurate of the actual method.

I completely agree this is good enough for all practical purposes. Updating the description sounds like a good idea, just to make sure there are no misunderstandings. Many thanks.

PR #163, opened 2020-08-06
Propagate exit status if given a child program to execute

chrisburr:
This more of a feature suggestion but it was simple enough for me to implement it as send a PR instead of opening an issue. I'm not sure if there are edge cases I haven't thought about it.

When passing an executable with arguments to would be useful to have the `prmon` process exit with the same error as the child, i.e.

Current behaviour:
```bash
$ prmon --json-summary test.json -- bash -c 'sleep 5 && exit 43'
Child process 894316 had non-zero return value: 43
$ echo $?
0
```
Suggested behaviour:
```bash
$ prmon --json-summary test.json -- bash -c 'sleep 5 && exit 43'
Child process 894316 had non-zero return value: 43
$ echo $?
43
```

Any thoughts? If backwards compatibility is important it could be hidden behind a command line flag (`--exit-with-child-code`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,10 @@ prmon [--pid PPP] [--filename prmon.txt] [--json-summary prmon.json] \
   and remaining arguments are passed to it; `prmon` will then monitor this process
   instead of being given a PID via `--pid`
 
+`prmon` will exit with `1` if there is a problem with inconsistent or 
+incomplete arguments. If `prmon` starts a program itself (using `--`) then
+`prmon` will exit with the exit code of the child process.
+
 
 ## Outputs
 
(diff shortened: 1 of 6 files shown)


graeme-a-stewart:
Hi @chrisburr - thanks a lot for this. I don't think I can see a use case for keeping the old 'always success' exit code in this case, so I think we don't really need the backwards compatibility flag.

Let me add a little note to the docs and add a unit test as well (our coverage is not great, but let's not make it worse for simple things!).

graeme-a-stewart:
@chrisburr I sent a PR to your branch with documentation and a test for this one

chrisburr:
@graeme-a-stewart Thanks for the patch to clean this up (I was on holiday at the time).

Can you make a new release so I can update `prmon` on conda-forge?

graeme-a-stewart:
Yes, will do - there's one other small thing I need to fix, but it will be done soon...

PR #166, opened 2020-09-03
Warn if monitored process walltime is very short

### LINKED ISSUE #165: Print a warning when runtime is < sample time
When the wallclock time of the monitored process is very low the values in the averages can be very unreliable, so we maybe need to print a warning in this case.

This was the source of #164.

graeme-a-stewart:
If the wallclock time of the monitored process is
very short then the averages calculated will not
be very reliable, due to sampling effects. At least
print a warning in that case.

Closes #165

### DIFF
--- a/package/src/prmon.cpp
+++ b/package/src/prmon.cpp
@@ -193,6 +193,15 @@ int ProcessMonitor(const pid_t mpid, const std::string filename,
   file << std::setw(2) << json_summary << std::endl;
   file.close();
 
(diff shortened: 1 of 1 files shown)


amete (approved):

amete:
Thanks a lot Graeme, this looks good.

Now that we're increasing the number of messages, I think it would be nice push #106. Currently, either turning on/off debugging or deciphering where the errors are actually coming from might be a bit challenging.

PR #168, opened 2020-09-04
Reformat and lint files with black and flake8

### LINKED ISSUE #157: Python code linter and formatter
Just as we started to standard format and lint our C++ code, we should do the same for Python.

I see no reason not to follow the [Scikit-HEP](https://scikit-hep.org/developer) advice on this and use [Black](https://black.readthedocs.io/en/latest/) for formatting.

For liniting, `pylint` seems to be stricter than `flake8`, but I don't have strong feelings (as we have few scripts then we could probably manage the strictness of `pylint`).

graeme-a-stewart:
A bit of an overdue clean-up of all of the Python scripts.

- All the scripts have been formatted with `black`
- All of the scripts have been checked against `flake8`

The scripts were also checked against `pylint`, and many issues from there fixed. However, it's exceedling strict and there are even some formatting decisions where `pylint` and `black` disagree, so that without specific exceptions added to the scripts the two checks can't agree. Most of the scripts test at 9/10 (or better) with `pylint`. We should endevour to keep the scripts good against `pylint` without enforcing it.

A GitHub action was added to run `flake8` as part of the CI.

Fixes #157


graeme-a-stewart:
Hi @amete - this should be good to go now.

All the scripts have been reformatted and small translation bugs are fixed
The new `flake8` test works fine, if you look at <https://github.com/HSF/prmon/runs/1082295343> then you'll see it failed for a small style violation

amete (approved):
Thanks a lot for this @graeme-a-stewart! The python code looks quite a bit cleaner now. All tests pass (locally as well). I ran the plotting script over a test file and the output looks good (i.e. no change). I'm merging now.

PR #172, opened 2020-10-20
Add Python data compression algorithm

rmaganza:
I implemented Andrea Sciabà's data compression algorithm in a new Python script as we had previously discussed and also added a new section in the README based on the visualization script one.

The algorithm could surely be expanded (i.e. by providing some information on the original series' variance) but it seems like a good starting point to me.

Script has been run through black and flake8.


amete (changes requested):
Hi @rmaganza, 

Thanks a lot for this PR and apologies for the belated review. I think we have a few things that need to be sorted out before we move ahead with this. 

Let me attach an example output from a production job ([memory_monitor_output.txt](https://github.com/HSF/prmon/files/5879669/memory_monitor_output.txt)). In the current implementation, it seems as if we're a bit too aggressive and having a bit of a problem interpreting the plateaus:

**Original: Text file size = 50 KB**
full

**Default: Text file size = 4.4 KB**
default

**Default + interpolate: Text file size = 15 KB**
interpolate

I think we should `interpolate` by default as it seems to provide a nice balance between the disk-space and the level of detail. A simple `tar` of the output reduces the size to 23 KB. So, there will still be a sizable gain in that configuration.

Many thanks.

Best,
Serhan

package/scripts/prmon_compress_output.py
```diff
@@ -0,0 +1,157 @@
...
+        default=0.1,
+        help="precision value for interpolation threshold",
+    )
+
+    parser.add_argument(
+        "--interpolate",
+        default=False,
+        action="store_true",
+        help="""Whether to interpolate the final obtained df,
+                or leave NAs for the different metrics""",
```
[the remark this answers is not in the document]
rmaganza: That's Pandas default behaviour when writing NAs to a text file. I think I could fix that fairly easily (e.g. by writing a NA string..)
I agree that interpolating by default is more informative, but I ran into some edge cases when testing originally for which actually the interpolation caused the file size to increase due to very high precision floating point numbers obtained during the interpolation, which does not seem very convenient by default. Perhaps I can add some rounding?

Let me know what you think.

README.md
```diff
@@ -151,6 +151,27 @@ The script allows the user to specify variables, their units, plotting
...
+The percentage threshold is set by the `--precision` parameter.
+* For more steady metrics: only keep the changepoints.
+
+The time index of the final output will be the union of the algorithm outputs of the single 
+time series. Each series will have a NA value where a point was deleted at a kept index, unless
+specified by the `--interpolate` parameter.
+
+Example:
+```sh
+prmon_compress_output.py --input prmon.txt --precision 0.3 --interpolate
```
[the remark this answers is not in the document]
rmaganza: First time I commit an executable script so I didn't know I had to commit permissions as well. Will add.

package/scripts/prmon_compress_output.py
```diff
@@ -0,0 +1,157 @@
...
+        "--output",
+        type=str,
+        default="prmon_compressed.txt",
+        help="name of the output compressed text file",
+    )
+
+    parser.add_argument(
+        "--precision",
+        type=lambda x: 0 < float(x) < 1
+        or parser.exit(-1, "Precision must be strictly between 0 and 1"),
```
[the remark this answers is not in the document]
rmaganza: You are indeed correct that I am passing `True` or `False` here. I must have messed up somewhere. 

Your solution works. 

This also works:  `type=lambda x: 0 < float(x) < 1 and float(x) or parser.exit(-1, "Precision must be strictly between 0 and 1")`.  Which was probably what I was aiming to.

Anyway I will fix this.

rmaganza:
Thank you for you suggestions @amete and don't worry about the timing. I added some comments to your correct remarks. 
Let me know what you think about the problem with the interpolation I mentioned.

Regards,
Serhan.

package/scripts/prmon_compress_output.py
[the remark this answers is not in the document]
amete: By default, the measurements in the text file are 64 bits unsigned integers. Therefore, I think you can round to the nearest integer without loosing too much information, at least for all intents and purposes.

amete:
Thanks a lot Riccardo. Please see above for my response and let me know if there are any loose ends.

README.md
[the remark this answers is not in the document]
amete: Sounds good, thanks.

package/scripts/prmon_compress_output.py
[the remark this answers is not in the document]
amete: Sounds good, thanks.

rmaganza:
Hi Serhan @amete,
the points we talked about should have been fixed.

Please take a last look and let me know if anything is off.

Regards,
Riccardo

README.md
```diff
@@ -151,6 +151,27 @@ The script allows the user to specify variables, their units, plotting
...
+
+### Data Compression
+
+The `prmon_compress_output.py` script (Python3) can be used to compress the output file
+while keeping the most relevant information.
+The compression algorithm works as follows, for each separate series: 
+* For rapidly changing metrics: for any three points A, B, and C, if the linear
+interpolation between A and C passes by B ± *threshold* * *range(metric)*, then B is deleted.
+The percentage threshold is set by the `--precision` parameter.
+* For more steady metrics: only keep the changepoints.
```
amete, line 163: Reading again, `rapidly changing` and `steady` metrics might be confusing. Looking at the script, the latter essentially contains the counts of processes, threads, and gpus while the former everything else. Maybe it's better to spell them out.

package/scripts/prmon_compress_output.py
```diff
@@ -0,0 +1,159 @@
...
+        default="prmon_compressed.txt",
+        help="name of the output compressed text file",
+    )
+
+    parser.add_argument(
+        "--precision",
+        type=lambda x: float(x)
+        if 0 < float(x) < 1
+        else parser.exit(-1, "Precision must be strictly between 0 and 1"),
+        default=0.1,
```
amete, line 126: I believe 0.05 is a better default, at least going over the example we have here:

**0.1 drops the file size from 50 KB to 15 KB, w/:**
reduced-new-default

**0.05 drops the file size from 50 KB to 16 KB, w/:**
reduced-new-0 05

Could you check and confirm? It looks like a price worth paying for 😄

amete (changes requested):
Thanks a lot @rmaganza. It looks like we're almost there. I have two minor comments. Then we should be good to go. Please let me know.

package/scripts/prmon_compress_output.py
rmaganza: This was a very open point, and it still is.

We also had some issues with deleting too much information on some test cases so I am happy to change the default.

README.md
rmaganza: Yes, naming is definitely off.

Would `MEMORY_IO_NETWORK_GPU_USAGE` and `NPROCS_NTHREADS_NGPUS` work better for you?

rmaganza: (I thought you were quoting the code)

I will fix this in the README for sure. I will also change the variable names in the code to avoid confusion.

rmaganza:
@amete Fixed last points in the README as well as the default precision value.

README.md
```diff
@@ -151,6 +151,29 @@ The script allows the user to specify variables, their units, plotting
...
+
+### Data Compression
+
+The `prmon_compress_output.py` script (Python3) can be used to compress the output file
+while keeping the most relevant information.
+The compression algorithm works as follows, for each separate series: 
+* For processes, threads and GPU counts: only keep the changepoints.
+* For rapidly changing metrics (i.e. the other variables): for any three points A, B, and C, if the linear
+interpolation between A and C passes by B ± *threshold* * *range(metric)*, then B is deleted.
+The percentage threshold is set by the `--precision` parameter.
```
amete, line 163: May I suggest something along the lines of:

```suggestion
The compression algorithm works as follows:
* For the number of processes, threads, and GPUs, only the measurements that are different with respect to the previous ones are kept.
* For all other metrics, only the measurements that satisfy an interpolation condition are kept.
This latter condition can be summarized as:
* For any three neighboring (and time-ordered) measurements, A, B, and C, B is deleted if the linear interpolation between A and C is consistent with B ± *threshold*. Otherwise, it's retained. The *threshold* can be configured via the `--precision` parameter (default: 0.05, i.e. 5%).
```

README.md
```diff
@@ -151,6 +151,29 @@ The script allows the user to specify variables, their units, plotting
...
+
+
+The time index of the final output will be the union of the algorithm outputs of the single 
+time series. Each series will have NA values where a point was deleted at a kept index and, unless otherwise
+specified by the `--skip-interpolate` parameter, will be linearly interpolated to maintain a consistent number of data points
```
amete, line 168: This one is about the code but when I use `--skip-interpolate` I still see missing entries and not `NA`. Is that expected?

package/scripts/prmon_compress_output.py
```diff
@@ -0,0 +1,159 @@
...
+    "read_bytes",
+    "write_bytes",
+    "rx_packets",
+    "tx_packets",
+    "rx_bytes",
+    "tx_bytes",
+    "gpufbmem",
+    "gpumempct",
+    "gpusmpct",
+]
```
amete, line 31: I noticed that we always drop the CPU measurements, namely `utime` and `stime`. I think we should also keep those.

amete (commented):
Just a few follow-ups/suggestions.

package/scripts/prmon_compress_output.py
rmaganza: Sure, we were dropping them in our testing back then but I agree they are useful for in-depth analyses.
Will add.

README.md
rmaganza: OK, I get what you mean now. 

1. Pandas writes NA values as empty strings when outputting to text file. I can change this to whatever you think works best.

2. On the other hand, I completely remove the line (sample at time *t*) if all the metrics were removed at time *t*. I could make it so that we retain the original number of lines and add empty lines in that case, if you think that's more useful.

rmaganza:
Thank you for your suggestions @amete.
I added a comment for the skip-interpolation case, which I think it's the only thing to iron out.

Regards,
Riccardo

README.md
rmaganza: Sure. Looks cleaner.

README.md
amete: I think that's OK to have empty strings. Pandas seem to correctly import the reduced data frame. I would only clarify the text above and say something like "dropped measurements are stored as empty strings, and when imported in pandas interpreted as NaN". Does that make sense?

rmaganza: Sure, I will add this to the README. The idea was anyway what you would re-import them into Pandas for the analysis.

rmaganza:
Hi @amete, 
I added the suggestions to the README.

I also added back the `utime` and `stime` variables. I appended them together with the CPU, GPU and thread numbers, since they work on a lower scale than the rest of the metrics and perhaps linearly interpolating would not be very suited in that case.

Let me know if that works for you.

amete:
Hi @rmaganza,

Could you quickly check the most up-to-date script along with the file I posted above? Locally, I see we get no size reduction anymore. We seem to drop only two rows, and replace all the metrics w/ the interpolated values instead. Am I missing something?

Best,
Serhan

rmaganza:
Hi @amete,
unfortunately, it seems like adding back `utime` and `stime` causes almost every row not to be deleted.
Every data point of the `utime` and `stime` series is marked as a changepoint and so it's not deleting any of the rows.

If you run the same code but remove `utime` and `stime` from the required metrics the output file size is 17 kb.

Let me know if you have any ideas about this, as I am not sure how we could keep them in the output file while maintaining the same pruning logic.

Regards,
Riccardo

amete:
Yes, `utime` and `stime` monotonically increase, typically at least. I think you can treat them as the other metrics, e.g. memory. This way you should be able to keep certain points where there are large enough changes but drop most of them. The other trick you can try is to take the first order derivative, i.e. `df['utime'].diff()` and use that to determine if a point should be dropped or not, instead of `df['utime']`. If you play around w/ these a little bit I'm sure you can find a good compromise but please let me know how it goes.

rmaganza:
Thanks, 
I was afraid interpolating a _time_ variable would not have had much sense logically, but you're right that moving them together with the memory variables does result in compression.

The test file size after compression is now 18kb.

amete (approved):
Thanks a lot @rmaganza. I think this looks good to go in as the first version. I think there are a few things that we can improve but given this is an optional script that's not going to affect the core functionality, let's follow-up on those later.

PR #174, opened 2020-10-27
Implement fix for missing i/o values

### LINKED ISSUE #173: Final values with 0 -> no IO values
Hi,

as discussed by email, here is the GitHub issue:

we've discussed the following prmon problem already a while ago and
thought it was fixed: final values set to 0 for the IO values in the
summary json file.

Here is a PanDA job of the recent 22.0.18 validation task:
https://bigpanda.cern.ch/job?pandaid=4858116576

The value of rchar, read_bytes, wchar, write_bytes are set to 0 in
memory_monitor_summary.json:
http://bigpanda.cern.ch//media/filebrowser/35939f9e-bbac-4f18-9978-8efb01e7db46/valid3/tarball_PandaJob_4858116576_LRZ-LMU_CLOUD/memory_monitor_summary.json

because the values of the last line in memory_monitor_output.txt are
mainly 0 in the last line:
https://bigpanda.cern.ch//media/filebrowser/35939f9e-bbac-4f18-9978-8efb01e7db46/valid3/tarball_PandaJob_4858116576_LRZ-LMU_CLOUD/memory_monitor_output.txt

The jobs use the latest stable prmon version setup by the PanDA pilot
from ALRB through: setupATLAS; lsetup prmon

It seems not to be site dependent for this task (happens at CERN-TO,
CERN, CERN-PROD_EOS, TW-FTT, LRZ-LMU_CLOUD, UNI_FREIBURG) and it happens
for 952 jobs out of 2054 (so it's pretty bad...).

For all these jobs, 0-values are reported back to PanDA and the
analytics ES cluster in Chicago, so we cannot do a full analysis w.r.t
to IO for many jobs - the individual job plotting is obviously
unaffected: https://bigpanda.cern.ch/prmonplots/4858116576/

Do you want me open a github issue ? Perhaps the logic in prmon has to
be changed for PanDA jobs since prmon is "killed" too late in the update
cycle ?

Cheers, Johannes

graeme-a-stewart:
Under some circumstances that are not very well
understood the process tree chain loses i/o values
(which would normally always be "inherited" by
parent processes as children exit).

To compensate for this the iomon module will now
record maximum i/o stats, and if the polled values
are below these it will replace the stats with the
previous maximums. A warning is also printed.

To test this logic there is a defined variable in
iomon.cpp that switches on code that artificially
provokes the problem, exercising the recovery
logic.

Closes #173

### DIFF
--- a/package/src/iomon.cpp
+++ b/package/src/iomon.cpp
@@ -9,6 +9,11 @@
 
 #include "utils.h"
 
+// Define this to be 1 to activate the artificial
+// supression of i/o values so that the recovery
+// of stats from the max values is checked
+#define IOMON_TEST 0
+
 // Constructor; uses RAII pattern to be valid
 // after construction
 iomon::iomon() : io_params{}, io_stats{} {
@@ -16,6 +21,7 @@ iomon::iomon() : io_params{}, io_stats{} {
   for (const auto& param : params) {
     io_params.push_back(param.get_name());
     io_stats[param.get_name()] = 0;
+    io_max_stats[param.get_name()] = 0;
   }
 }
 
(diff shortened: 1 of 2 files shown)


amete (approved):
Thanks a lot @graeme-a-stewart. I don't see any issue with this patch as the disk I/O metrics should really be monotonically increasing. It's technically working as expected, too. Having said that, the cause of the issue in #173 is still not clear to me. In any case, this is a fair workaround.

PR #178, opened 2020-11-20
Implemented simple monitor disabling

### LINKED ISSUE #107: Options to enable/disable monitoring components

graeme-a-stewart:
Added new switch, --disable/-d that allows
monitors to be disabled. Switch can be given
multiple times to disable several components.

Validity of component name is checked and
walllmon cannot be disabled.

Added a new test to ensure that the switch
fucntions correctly.

README.md and help text updated.

Closes #107


graeme-a-stewart:
There seems to be an issue with our build actions due to a GH security fix. I'll follow up on that.

graeme-a-stewart:
I think the codefactor warning can be ignored in this case (it's what we used for ages and injecting a bogus "pstree" into the PATH would just cause prmon to break.

amete (changes requested):
Thanks a lot @graeme-a-stewart. The new functionality looks good to me. I have just a few minor comments.

amete:
I resolved the simple conflict through the web interface and the associated `CI` passed (although not posted above due to `clang-format` interfering):

https://github.com/HSF/prmon/actions/runs/378656295

We agree to ignore the `CodeFactor` error for the time being. Therefore, I'm merging this now. Many thanks @graeme-a-stewart.

PR #181, opened 2021-01-20
Fix netstats rollover

### LINKED ISSUE #180: Overflowing network metrics
It's reported that under certain circumstances, network metrics seem to be overflowing:

```
Time	[...]	rx_bytes	rx_packets	tx_bytes	tx_packets
[...]
1607450732	[...]	182018452734	148438951	4801778623	13500137
1607450793	[...]	18446743984687580690	18446744073623414305	18446744072751791668	18446744073703405569
[...]
```

The problem is reported for the following ATLAS PanDA task: https://bigpanda.cern.ch/job?pandaid=4914743947

The full `prmon` text file result can be found [here](https://bigpanda.cern.ch/media/filebrowser/bd2f7182-6a2a-4302-b899-92838ecf6bb2/mc15_13TeV/tarball_PandaJob_4914743947_SWT2_CPB/memory_monitor_output.txt) and is also attached at [1].

[1] [memory_monitor_output.txt.tar.gz](https://github.com/HSF/prmon/files/5665527/memory_monitor_output.txt.tar.gz)

graeme-a-stewart:
Change the way that network statistics are calculated so that if any stat rolls over (64bit uint overflow or another issue with the network counters), such that during an iteration
```
current_value < previous_value
```
Then the change is ignored for this cycle and a warning is printed.

Fixed a bug in the network stats test scripts (target values not updated correctly if `--blocks` was specified).

Fixed a few minor flake8 issues in test scripts.

Closes #180

### DIFF
--- a/package/src/netmon.cpp
+++ b/package/src/netmon.cpp
@@ -7,7 +7,6 @@
 #include <time.h>
 
 #include <cstring>
-#include <limits>
 #include <memory>
 
 #include "utils.h"
--- a/package/src/netmon.cpp
+++ b/package/src/netmon.cpp
@@ -7,6 +7,7 @@
 #include <time.h>
 
 #include <cstring>
+#include <limits>
 #include <memory>
 
 #include "utils.h"
@@ -27,8 +28,10 @@ netmon::netmon(std::vector<std::string> netdevs)
   open_interface_streams();
 
   // Ensure internal stat counters are initialised properly
-  read_raw_network_stats(network_stats_start);
-  read_raw_network_stats(network_stats);
+  read_raw_network_stats(network_stats_last);
+  for (const auto& i : network_stats_last) {
+    network_net_counters[i.first] = 0;
+  }
 }
 
 // Get all available network devices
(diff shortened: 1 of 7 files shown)


amete (approved):
Thanks a lot @graeme-a-stewart. The updates look good to me. Instead of doing `now-start`, we now do `now-last` and increment the counters accordingly, except when there is a drop. 

The only bit I think can be improved, if we want to monitor how often the check fails, is to keep an internal counter, say `network_error_counters`, increment this when we bump into a decrease (in addition to the error message), and push this into the JSON at the end of the job. The error message itself will probably be fairly hard to fish out in production jobs.

Having said that, I don't really feel too strongly on that. Therefore, let me approve as is and let you decide 😄

graeme-a-stewart:
Thanks for taking a look @amete. It would be interesting to think about error counters, but I would prefer that we added it more generally to the `Imonitor` class and that all of the monitors could use it. Otherwise we get into a situation where we're handling a monitor in a special way, which compromises the design of the code.

So my inclination is to accept this PR and create a new issue if we want to handle errors in a more systematic way (also, there was my thought to introduce more formal logging in general, #106).

PR #183, opened 2021-01-27
Implement monitor disable via environment

### LINKED ISSUE #182: Disable monitors more flexibly
Although we implemented a mechanism to disable specific monitors from the command line (#178) it doesn't help much if a lower part of the workflow is the one in charge of launching `prmon`, e.g., the transform infrastructure in Athena.

For these cases it would be nice to also support disabling monitors "from the outside". This could be done by looking for an enviroment variable that would trigger disabling one or more monitiors. Then users can set this in their environment and subsequent invocations of `prmon` will respect it.

graeme-a-stewart:
Allow monitors to be disabled via the `PRMON_DISABLE_MONITOR` environment variable.

Closes #182

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,20 @@ incomplete arguments. If `prmon` starts a program itself (using `--`) then
 When invoked with `-h` or `--help` usage information is printed, as well as a
 list of all available monitoring components.
 
+### Environment Variables
+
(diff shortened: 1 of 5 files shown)


amete (approved):
LGTM! Thanks a lot @graeme-a-stewart.

Issue #189, opened 2021-06-14
spdlog vs boost.log for prmon logging


quantum-shift:
It was decided to test spdlog and boost.log for integration with prmon. spdlog turned out to be a clear winner in all tests. 

The speeds of spdlog and boost.log were compared. Three kinds of tests were conducted varying the number and types of sinks being logged into:
a) Only a console sink
b) Only a file sink
c) Both console and file sinks.

1e6 messages were logged to the sinks and the execution times were recorded. Each test was repeated 5 times.

Here are the results:

|          |             | ONLY CONSOLE SINK |        |  |             | ONLY FILE SINK |        |  |             | BOTH CONSOLE AND FILE SINKS |        |
|----------|-------------|-------------------|--------|--|-------------|----------------|--------|--|-------------|-----------------------------|--------|
|          | Time (in s) | Average           | Median |  | Time (in s) | Average        | Median |  | Time (in s) | Average                     | Median |
|          |             |                   |        |  |             |                |        |  |             |                             |        |
|          | 4.431       |                   |        |  | 0.125       |                |        |  | 4.403       |                             |        |
|          | 4.364       |                   |        |  | 0.111       |                |        |  | 4.459       |                             |        |
| SPDLOG   | 4.35        | 4.3892            | 4.35   |  | 0.136       | 0.121          | 0.119  |  | 4.531       | 4.4984                      | 4.522  |
|          | 4.459       |                   |        |  | 0.114       |                |        |  | 4.522       |                             |        |
|          | 4.342       |                   |        |  | 0.119       |                |        |  | 4.577       |                             |        |
|          |             |                   |        |  |             |                |        |  |             |                             |        |
|          | 7.429       |                   |        |  | 2.812       |                |        |  | 9.921       |                             |        |
|          | 7.271       |                   |        |  | 2.878       |                |        |  | 10.235      |                             |        |
| BOOSTLOG | 7.28        | 7.2894            | 7.28   |  | 2.949       | 2.8996         | 2.914  |  | 9.83        | 9.8566                      | 9.83   |
|          | 7.298       |                   |        |  | 2.914       |                |        |  | 9.579       |                             |        |
|          | 7.169       |                   |        |  | 2.945       |                |        |  | 9.718       |                             |        |

Some other differences:
a) It was much harder to install and configure boost.log than spdlog.
b) The number of header files to include for spdlog was much less than that for boost.log to accomplish the same behaviour.
c) The amount of code required to initialise sinks and logger to produce similar output for

[...]

PR #190, opened 2021-06-24
Added logging functionality (spdlog)

quantum-shift:
As discussed in issue #189, spdlog is integrated into prmon. spdlog (version 1.8.5) is added as a submodule. All previous logging statements are replaced with the new logger.

Further, command line switches will be added to handle the logging level of monitors.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,11 @@ endif()
 set(PROFILE_GPERFTOOLS "${PROFILE_GPERFTOOLS}"
   CACHE BOOL "Build with the GPerfTools profiler library" FORCE)
 
+if(NOT BUILD_BENCHMARK_LOG)
+  set(BUILD_BENCHMARK_LOG OFF)
+endif()
+set(BUILD_BENCHMARK_LOG "${BUILD_BENCHMARK_LOG}"
+  CACHE BOOL "Build binary that benchmarks spdlog speed" FORCE)
 
 # Define a default build type when using a single-mode tool like make/ninja
 # We make this default to Release, unless profiling is enabled, in which
@@ -92,6 +97,11 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE
 #--- project specific subdirectories -------------------------------------------
 add_subdirectory(package)
 
+#--- include directories -------------------------------------------
+target_include_directories(prmon PRIVATE ${PROJECT_SOURCE_DIR}/package/include)
+if (BUILD_BENCHMARK_LOG)
+  target_include_directories(benchmark-log PRIVATE ${PROJECT_SOURCE_DIR}/package/include)
+endif(BUILD_BENCHMARK_LOG)
 #--- create uninstall target ---------------------------------------------------
 include(cmake/prmonUninstall.cmake)
 
--- a/package/CMakeLists.txt
+++ b/package/CMakeLists.txt
@@ -10,8 +10,11 @@ add_executable(prmon src/prmon.cpp
            src/wallmon.cpp
            src/memmon.cpp
            src/nvidiamon.cpp
+           src/MessageBase.cpp
            )
-
+if (BUILD_BENCHMARK_LOG)
+  add_executable(benchmark-log src/benchmark.cpp)
+endif(BUILD_BENCHMARK_LOG)
 # Some older versions of the nlohmann_json package don't define
 # this target (< 3.0.0 ?). The build on these platforms seems to
 # go ok even without this "link library" definition.
--- a/package/src/MessageBase.h
+++ b/package/src/MessageBase.h
@@ -1,0 +1,31 @@
+#ifndef PRMON_MESSAGEBASE_H
+#define PRMON_MESSAGEBASE_H 1
+
+#include "spdlog/sinks/basic_file_sink.h"
+#include "spdlog/sinks/stdout_color_sinks.h"
+#include "spdlog/spdlog.h"
+
+// Global sinks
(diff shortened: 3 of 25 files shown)


CMakeLists.txt
```diff
@@ -92,9 +97,15 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE
...
+if (BUILD_BENCHMARK_LOG)
+  target_include_directories(benchmark-log PRIVATE ${PROJECT_SOURCE_DIR}/package/include)
+endif(BUILD_BENCHMARK_LOG)
 #--- create uninstall target ---------------------------------------------------
 include(cmake/prmonUninstall.cmake)
 
 #--- code format targets -------------------------------------------------------
 include(cmake/clang-format.cmake)
 include(cmake/python-format.cmake)
+
```
amete, line 111: We probably don't need this empty line.

package/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 find_package(nlohmann_json REQUIRED)
 
+
```
amete, line 3: We probably don't need this empty line.

package/src/MessageBase.h
```diff
@@ -0,0 +1,31 @@
...
+
+ public:
+  void set_log_level(const spdlog::level::level_enum& level);
+  void debug(const std::string& message);
+  void info(const std::string& message);
+  void warning(const std::string& message);
+  void error(const std::string& message);
+};
+
+#endif  // PRMON_MESSAGEBASE_H
```
amete, line 31: Missing newline at the end of the line.

package/src/countmon.cpp
```diff
@@ -10,6 +10,7 @@
 #include <iostream>
 #include <sstream>
 
+#include "MessageBase.h"
```
amete, line 13: We don't really need to include this header in the source files since it's already in the header, no?

package/src/countmon.h
```diff
@@ -6,17 +6,19 @@
 
 #ifndef PRMON_COUNTMON_H
 #define PRMON_COUNTMON_H 1
+#define MY_NAME "countmon"
```
amete, line 9: I think we can define this in the source file instead (applies to other cases too) and un-define it after using. Also, we can perhaps call it something more explanatory, e.g. `MONITOR_NAME`.

package/src/utils.cpp
```diff
@@ -103,4 +103,4 @@ const void prmon::fill_units(nlohmann::json& unit_json,
       unit_json["Units"]["Avg"][param.get_name()] = param.get_avg_unit();
   }
   return;
-}
+}
```
amete, line 106: Missing newline at the end of the line.

amete (changes requested):
Hi @quantum-shift,

Thanks a lot for the PR! Overall, everything looks fairly nice. I noticed that we still have a few uses of `std::cout` e.g. https://github.com/quantum-shift/prmon/blob/f036e9dede4d5f0329746d9ea4cbd5d3921d1d37/package/src/prmon.cpp#L307 etc. I think it would be good to unify these as well. I have some more detailed comments below. Please let me know, many thanks.

Best,
Serhan

quantum-shift:
Thank you, @amete. It was decided to not change the help output of prmon (that is, not integrate it with the logger). Hence, I have kept the std::cout statements there. Please let me know if there was a change in plan. I have addressed your other comments in a commit.

package/src/countmon.h
graeme-a-stewart: If it's going to be private maybe should it be something like `THIS_MONITOR_NAME_` (or were you thinking of calling it, e.g., `COUNTMON_NAME_` (different in each monitor).

I agree that undefining it at the end is smart - stops it from leaking out.

graeme-a-stewart:
> 
> 
> Thank you, @amete. It was decided to not change the help output of prmon (that is, not integrate it with the logger). Hence, I have kept the std::cout statements there. Please let me know if there was a change in plan. I have addressed your other comments in a commit.

Yes, this also my recollection (it's quite standard behaviour for `--help``, where one would certainly not want it to go to a logging file!).

amete:
> > Thank you, @amete. It was decided to not change the help output of prmon (that is, not integrate it with the logger). Hence, I have kept the std::cout statements there. Please let me know if there was a change in plan. I have addressed your other comments in a commit.
> 
> Yes, this also my recollection (it's quite standard behaviour for `--help`, where one would certainly not want it to go to a logging file!).

That indeed makes sense 👍

CMakeLists.txt
```diff
@@ -29,6 +29,11 @@ endif()
 set(PROFILE_GPERFTOOLS "${PROFILE_GPERFTOOLS}"
   CACHE BOOL "Build with the GPerfTools profiler library" FORCE)
 
+if(NOT BUILD_BENCHMARK_LOG)
+  set(BUILD_BENCHMARK_LOG OFF)
+endif()
+set(BUILD_BENCHMARK_LOG "${BUILD_BENCHMARK_LOG}"
+  CACHE BOOL "Build binary which benchmarks spdlog speed" FORCE)
```
graeme-a-stewart, line 36: Very trivial point, but can you change `which` to `that` (it's a defining clause)

graeme-a-stewart (changes requested):
Just one utterly trivial point from me! Thanks @quantum-shift

graeme-a-stewart (approved):

amete (approved):
LGTM, thanks a lot @quantum-shift!

Issue #191, opened 2021-06-29
Fix compilation issues under gcc11 


amete:
As reported by Johannes, we seem to have issues building in `gcc11`:

```
/build/atnight/localbuilds/nightlies/master--dev4LCG/build/build/AthenaExternals/src/prmon/package/src/registry.h: In instantiation of 'static std::vector<std::__cxx11::basic_string<char> > registry::Registry<T, Pack>::list_registered() [with T = Imonitor; Pack = {}]':
/build/atnight/localbuilds/nightlies/master--dev4LCG/build/build/AthenaExternals/src/prmon/package/src/prmon.cpp:42:60:   required from here
/build/atnight/localbuilds/nightlies/master--dev4LCG/build/build/AthenaExternals/src/prmon/package/src/registry.h:68:21: error: loop variable 'k' creates a copy from type 'const std::pair<const std::__cxx11::basic_string<char>, std::function<Imonitor*()> >' [-Werror=range-loop-construct]

   68 |     for (const auto k : ctors()) {
      |                     ^
/build/atnight/localbuilds/nightlies/master--dev4LCG/build/build/AthenaExternals/src/prmon/package/src/registry.h:68:21: note: use reference type to prevent copying
   68 |     for (const auto k : ctors()) {
      |                     ^
      |                     &
```

We should go ahead and make sure we can compile w/ `gcc11` (it would be nice to extend the CI coverage too).

graeme-a-stewart:
Closed with #192

PR #192, opened 2021-06-29
Fix for #191

amete:
This one is fairly simple. What we currently have seems to trigger an `-Werror=range-loop-construct` on `gcc11` [1]. Now we simply use a reference to avoid copying.

[1] I took `gcc11` from `source /cvmfs/sft.cern.ch/lcg/views/dev4/latest/x86_64-centos7-gcc11-dbg/setup.sh` but we can actually setup a dedicate CI for this.

### DIFF
--- a/package/src/registry.h
+++ b/package/src/registry.h
@@ -65,7 +65,7 @@ class Registry {
 
   static std::vector<std::string> list_registered() {
     std::vector<std::string> list{};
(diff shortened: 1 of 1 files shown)


graeme-a-stewart (approved):
Ah yes, good catch from the compiler. Thanks for the fix @amete!

PR #194, opened 2021-07-19
Fix: cpumon doesn't reset values each iteration

quantum-shift:
The iterating variable here: https://github.com/HSF/prmon/blob/9e762bc589de6229601b310e4be1adc4e3703d5b/package/src/cpumon.cpp#L31 isn't a reference. Hence, the values remain unchanged and are not reset every iteration.


amete (approved):
Very good catch @quantum-shift!

graeme-a-stewart (approved):

PR #195, opened 2021-07-27
Precooked tests for monitors

quantum-shift:
Monitors have been adapted to optionally read from precooked sources. Monotonicity tests have been added for cpumon, iomon, netmon and memmon. A value test has been added for nvidiamon.

### Changes:

- The update_stats() methods of the abstract Imonitor base class and all of its derived classes have been modified to accept an optional argument read_path that points to the directory the monitor should read from.
- Auxiliary methods have been added to netmon and nvidiamon classes for reading from precooked sources to preserve the structure of the update_stats() method of these classes since both these classes use separate functions to procure the data.

### Source files added:
- scripts/precook_tests.py: This generates the scripts/precooked_tests directory that contains the precooked test data.
- tests/test_values.cpp: This is a test harness containing the new tests. GoogleTest framework is used for the tests. Presently, it contains monotonicity tests for cpumon, iomon, netmon and memmon. The cpumon and memmon tests fail as protection against dropping values has not been added yet.

PR #196, opened 2021-08-02
Added new montored value class

### LINKED ISSUE #185: Protect monotonic increasing stats against race conditions
@rmaganza reports an unusual case where the `utime` measured by `prmon` suffered an anomolous drop, right at the end of the job (PandaID 4929654149).

There is a possible race condition that might cause this:

1. prmon polls the list of PIDs owned by the mother process
2. prmon starts to loop over the process list and collects stats from the mother
3. one or more children exit before they can be polled
4. although the kernel now attributes child resource consumption back to the mother we don’t see it, because the mother was already polled
5. polling the stats for the exited processes (obviously) fails

We should protect monotonic increasing stats against dropping,

graeme-a-stewart:
This is a big PR, as it introduces a specific class that handles monitored variables, thus refactoring of all monitors to use this class has been added. Also added are namespace `using` statements that define properly the types used by monitors.

The big advantages of this are:
- The intelligent handling of monitored values becomes part of the class, thus less specific code to catch errors needs to be added in each monitor itself
- Specifically values can be flagged as *monotonic*, which protects them against ever decreasing
- Data members and return types now use properly namespaced definitions

This enabled specific monitor protections for decreased values to be removed from a few monitors (iomon, netmon) as well as adding protection for CPU and walltime monitors.

Note that the class does not yet use loggers, but that this should be added (e.g., as a pointer to the logger instance, with `nullptr` meaning fallback to `std::cerr`).

Closes #185

### DIFF
--- a/package/src/iomon.cpp
+++ b/package/src/iomon.cpp
@@ -22,7 +22,8 @@ iomon::iomon() : io_stats{} {
   log_init(MONITOR_NAME);
 #undef MONITOR_NAME
   for (const auto& param : params) {
-    io_stats.emplace(std::make_pair(param.get_name(), prmon::monitored_value(param, true)));
+    io_stats.emplace(
+        std::make_pair(param.get_name(), prmon::monitored_value(param, true)));
   }
 }
 
@@ -88,7 +89,8 @@ prmon::monitored_average_map const iomon::get_json_average_stats(
   prmon::monitored_average_map io_average_stats{};
   for (const auto& io_param : io_stats) {
     io_average_stats[io_param.first] =
-        double(io_param.second.get_value() * sysconf(_SC_CLK_TCK)) / elapsed_clock_ticks;
+        double(io_param.second.get_value() * sysconf(_SC_CLK_TCK)) /
+        elapsed_clock_ticks;
   }
   return io_average_stats;
 }
--- a/package/src/iomon.cpp
+++ b/package/src/iomon.cpp
@@ -88,7 +88,7 @@ prmon::monitored_average_map const iomon::get_json_average_stats(
   prmon::monitored_average_map io_average_stats{};
   for (const auto& io_param : io_stats) {
     io_average_stats[io_param.first] =
-        double(io_param.second.get_value() * sysconf(_SC_CLK_TCK)) /
+        prmon::avg_value(io_param.second.get_value() * sysconf(_SC_CLK_TCK)) /
         elapsed_clock_ticks;
   }
   return io_average_stats;
--- a/package/src/iomon.cpp
+++ b/package/src/iomon.cpp
@@ -22,8 +22,7 @@ iomon::iomon() : io_stats{} {
   log_init(MONITOR_NAME);
 #undef MONITOR_NAME
   for (const auto& param : params) {
-    io_stats.emplace(
-        std::make_pair(param.get_name(), prmon::monitored_value(param, true)));
+    io_stats.emplace(param.get_name(), prmon::monitored_value(param, true));
   }
 }
 
--- a/package/src/iomon.cpp
+++ b/package/src/iomon.cpp
@@ -18,22 +18,22 @@
 
 // Constructor; uses RAII pattern to be valid
 // after construction
-iomon::iomon() : io_params{}, io_stats{} {
+iomon::iomon() : io_stats{} {
   log_init(MONITOR_NAME);
 #undef MONITOR_NAME
-  io_params.reserve(params.size());
   for (const auto& param : params) {
-    io_params.push_back(param.get_name());
-    io_stats[param.get_name()] = 0;
-    io_max_stats[param.get_name()] = 0;
+    io_stats.emplace(std::make_pair(param.get_name(), prmon::monitored_value(param, true)));
   }
 }
 
 void iomon::update_stats(const std::vector<pid_t>& pids) {
+  prmon::monitored_value_map io_stat_update{};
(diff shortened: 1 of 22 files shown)


amete (changes requested):
Thanks a lot for this PR @graeme-a-stewart. In general it looks good. I have only a few minor comments below.

By the way, locally testing your changes I bumped into a failed unit test:

```
12/14 Test #12: testUnits ........................***Failed    3.82 sec
```

with the output:

```
12/14 Testing: testUnits
12/14 Test: testUnits
Command: "/cvmfs/sft-nightlies.cern.ch/lcg/views/dev4/Mon/x86_64-centos7-gcc11-dbg/bin/python3" "test_cpu.py" "--units" "--time" "3" "--slack" "0" "--interval" "1" 
Directory: /afs/cern.ch/user/a/amete/workspace/prmon-workspace/build/package/tests
"testUnits" start time: Aug 02 16:06 CEST
Output:
----------------------------------------------------------
Will run for 3s using 1 process(es) and 1 thread(s)
[2021-08-02 16:06:17.109] [prmon] [warning] Wallclock time of monitored process was less than the monitoring interval, so average statistics will be unreliable
F
======================================================================
FAIL: test_run_test_with_params (__main__.setup_configurable_test.<locals>.ConfigurableProcessMonitor)
Actual test runner
----------------------------------------------------------------------
  File "/afs/cern.ch/work/a/amete/prmon-workspace/build/package/tests/test_cpu.py", line 82, in test_run_test_with_params
    self.assertLessEqual(
AssertionError: 4 not less than or equal to 3.0 : Too high value for wall time (expected maximum of 3.0, got 4)

----------------------------------------------------------------------
Ran 1 test in 3.724s

FAILED (failures=1)
<end of output>
Test time =   3.82 sec 
----------------------------------------------------------
Test Failed.
"testUnits" end time: Aug 02 16:06 CEST
"testUnits" time elapsed: 00:00:03
----------------------------------------------------------
```

The test succeeded on the second try so we might need to take a closer look at the tests to make them more robust but that's for another issue/PR.

package/src/netmon.cpp
```diff
@@ -121,12 +120,13 @@ std::map<std::string, unsigned long long> const netmon::get_json_total_stats() {
 }
```
quantum-shift, line 120: The return type can be ```prmon::monitored_value_map``` as for other monitors.

package/src/iomon.cpp
```diff
@@ -42,65 +43,56 @@ void iomon::update_stats(const std::vector<pid_t>& pids) {
...
-std::map<std::string, double> const iomon::get_json_average_stats(
+prmon::monitored_average_map const iomon::get_json_average_stats(
     unsigned long long elapsed_clock_ticks) {
-  std::map<std::string, double> json_average_stats{};
+  prmon::monitored_average_map io_average_stats{};
   for (const auto& io_param : io_stats) {
-    json_average_stats[io_param.first] =
-        double(io_param.second * sysconf(_SC_CLK_TCK)) / elapsed_clock_ticks;
+    io_average_stats[io_param.first] =
+        double(io_param.second.get_value() * sysconf(_SC_CLK_TCK)) /
```
quantum-shift, line 92: We can change double to prmon::avg_value for consistency.

package/src/netmon.cpp
```diff
@@ -121,12 +120,13 @@ std::map<std::string, unsigned long long> const netmon::get_json_total_stats() {
...
+prmon::monitored_average_map const netmon::get_json_average_stats(
     unsigned long long elapsed_clock_ticks) {
-  std::map<std::string, double> json_average_stats{};
-  for (auto& stat : network_net_counters) {
-    json_average_stats[stat.first] =
-        double(stat.second * sysconf(_SC_CLK_TCK)) / elapsed_clock_ticks;
+  prmon::monitored_average_map json_average_stats{};
+  for (auto& value : net_stats) {
+    json_average_stats[value.first] =
+        double(value.second.get_value() * sysconf(_SC_CLK_TCK)) /
```
quantum-shift, line 128: We can change double to prmon::avg_value for consistency.

quantum-shift (changes requested):
Thank you, @graeme-a-stewart. 
I am curious if there may arise a situation in the future where we have to add a metric to a monitor so that we have both monotonic and non-monotonic stats for the same monitor?

graeme-a-stewart:
Latest changes should address all of the review comments - thank you both for them

Still TODO: 
- Documentation update
- Unit tests

package/src/parameter.h
```diff
@@ -31,6 +39,63 @@ class parameter {
...
 using parameter_list = std::vector<parameter>;
+
+class monitored_value {
+  // monitored_value class gives an interface to a value that we monitor
+  // in prmon
+  // It holds metadata about the value (name, units - using the parameter class)
+  // as well as internal counters and accessors. In particular the monotonic
+  // flag will prevent the measured value from being lowered.
+ private:
+  const parameter param;
```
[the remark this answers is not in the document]
graeme-a-stewart: I think this is sort of an ATLAS convention (I couldn't find anything in <http://isocpp.github.io/CppCoreGuidelines>), but fine with me.

quantum-shift (approved):

graeme-a-stewart:
OK, all PR comments have been addressed and last issues fixed. Unit tests are running correctly for the new classes as well, thanks to the Google Test setup in #197.

Will merge this so that @quantum-shift can work on rebasing #195.

Issue #201, opened 2021-09-15
Cleanup CI


graeme-a-stewart:
As discussed, prior to the 3.0 release we will clean up the CI:

- Drop U18 as a target
- Drop Python2 support
- Add a build against a more modern gcc
- Add a build against clang

graeme-a-stewart:
I need to have a look on my other machine where I have docker available. The HSF images I build from the dockerfiles here:

<https://github.com/graeme-a-stewart/hithe>

So it should be easy to ensure that we have a fresh `g++` and `clang++` in one of those containers. `build-test.sh` should just get passed the correct compilers (with an `scl enable` prefix, if needed on CS8).

graeme-a-stewart:
I had a look at running CI tests with clang. It will not be hard, but will need a bit of manipilation so that, e.g., platform `c8-dev:clang` could be split properly into the container image and a compilation option.

PR #203, opened 2021-09-21
Drop python2 support and remove u18 from the supported platforms

amete:
This PR drops the `python2` support and removes `u18` from the supported platforms. It partly addresses #201. Adding a build against a more modern `gcc` and `clang` in the `CI` will be included in a separate PR.


graeme-a-stewart (changes requested):
Looks great, but I think in the README we should mention somewhere that we require Python3 (>=3.6), maybe specifically for running the tests

graeme-a-stewart (approved):

PR #211, opened 2021-12-12
Fix the clang compilation issue reported in #210

### LINKED ISSUE #210: prmon v3.0.0 compilation problem with clang10/13
Hi,

the prmon version 3.0.0 has a compilation problem with clang10 and clamg13 - when using clang13 from `asetup Athena,master,latest,clang13`, the prmon compilation fails:
```
[..]
In file included from /afs/cern.ch/user/e/elmsheus/work/testarea/prmon/prmon/package/src/wallmon.cpp:4:
/afs/cern.ch/user/e/elmsheus/work/testarea/prmon/prmon/package/src/wallmon.h:30:22: error: private field 'start_time_clock_t' is not used [-Werror,-Wunused-private-field]
  unsigned long long start_time_clock_t, current_clock_t;
                     ^
/afs/cern.ch/user/e/elmsheus/work/testarea/prmon/prmon/package/src/wallmon.h:30:42: error: private field 'current_clock_t' is not used [-Werror,-Wunused-private-field]
  unsigned long long start_time_clock_t, current_clock_t;
                                         ^
2 errors generated.
make[2]: *** [package/CMakeFiles/prmon.dir/src/wallmon.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[ 96%] Building CXX object package/CMakeFiles/prmon.dir/src/nvidiamon.cpp.o
make[1]: *** [package/CMakeFiles/prmon.dir/all] Error 2
make: *** [all] Error 2
```

This is also visible in the latest Athena external build logs after the upgrade to atlasexternals 2.0.123 which ships prmon 3.0.0: https://atlas-computing.web.cern.ch/atlas-computing/links/distDirectory/gitwww/MASTERWebArea/ardoc_web_areamaster64BC7AthenaOpt/ARDOC_Log_2021-12-11T2101/ardoc_externals_build.html

Cheers, Johannes

amete:
This should fix the reported error. However, I also noticed that the memory unit tests are failing in the clang builds. That needs to be investigated (might be related to different optimizations). A good reminder to add a clang based CI workflow to catch similar issues in the future (as discussed in #201). 

Closes #210

### DIFF
--- a/package/src/wallmon.cpp
+++ b/package/src/wallmon.cpp
@@ -30,7 +30,7 @@ std::pair<int, unsigned long long> wallmon::get_mother_starttime(
   std::vector<std::string> stat_entries{};
   stat_entries.reserve(52);
   std::string tmp_str{};
-  unsigned long long start_time_clock_t;
+  unsigned long long start_time_clock;
 
   std::stringstream stat_fname{};
   stat_fname << "/proc/" << mother_pid << "/stat" << std::ends;
@@ -40,7 +40,7 @@ std::pair<int, unsigned long long> wallmon::get_mother_starttime(
     if (proc_stat) stat_entries.push_back(tmp_str);
   }
   if (stat_entries.size() > prmon::uptime_pos) {
-    start_time_clock_t = std::stol(stat_entries[prmon::uptime_pos]);
+    start_time_clock = std::stol(stat_entries[prmon::uptime_pos]);
   } else {
     // Some error happened!
     std::stringstream strm;
@@ -51,7 +51,7 @@ std::pair<int, unsigned long long> wallmon::get_mother_starttime(
     warning(strm.str());
     return std::pair<int, unsigned long long>{1, 0L};
   }
-  return std::pair<int, unsigned long long>{0, start_time_clock_t};
+  return std::pair<int, unsigned long long>{0, start_time_clock};
 }
 
 void wallmon::update_stats(const std::vector<pid_t>& pids,
--- a/package/src/wallmon.h
+++ b/package/src/wallmon.h
@@ -27,8 +27,6 @@ class wallmon final : public Imonitor, public MessageBase {
   // "map" of monitored parameters (even if there's only one!)
   prmon::monitored_list walltime_stats;
 
-  unsigned long long start_time_clock_t, current_clock_t;
-
   // Only need to get the mother start time once, so use
   // a bool to say when it's done
   bool got_mother_starttime;


amete:
BTW, just to follow up, the memory unit tests succeeds upon explicitly disabling clang's `malloc` optimizations via `-fno-builtin`. So, we might need to tweak `mem-burner` a bit.

package/src/wallmon.cpp
```diff
@@ -30,7 +30,7 @@ std::pair<int, unsigned long long> wallmon::get_mother_starttime(
   std::vector<std::string> stat_entries{};
   stat_entries.reserve(52);
   std::string tmp_str{};
-  unsigned long long start_time_clock_t;
+  unsigned long long start_time_clock;
```
graeme-a-stewart, line 33: I think this is an uncecessary change

graeme-a-stewart (changes requested):

package/src/wallmon.cpp
amete: That's indeed not necessary. I just renamed the variable as typically typedefs get an `_t` suffix but that's purely stylistic. Let me revert this.

graeme-a-stewart:
> BTW, just to follow up, the memory unit tests succeed upon explicitly disabling clang's `malloc` optimizations via `-fno-builtin`. So, we might need to tweak `mem-burner` a bit but that's not directly relevant to this PR.

Yeah, it looks like somehow clang manages to use less VMEM that the standard malloc does, which trips the test.

We should either fix the test expectations or add that `-fno-builtin` flag - I wouldn't like to have to fix clang issues twice

amete:
> > BTW, just to follow up, the memory unit tests succeed upon explicitly disabling clang's `malloc` optimizations via `-fno-builtin`. So, we might need to tweak `mem-burner` a bit but that's not directly relevant to this PR.
> 
> Yeah, it looks like somehow clang manages to use less VMEM that the standard malloc does, which trips the test.
> 
> We should either fix the test expectations or add that `-fno-builtin` flag - I wouldn't like to have to fix clang issues twice

We can compile `mem-burner` without the relevant optimization, which seems to do the trick, at least locally:

`diff --git a/package/tests/CMakeLists.txt b/package/tests/CMakeLists.txt
index b63d834..d348112 100644
--- a/package/tests/CMakeLists.txt
+++ b/package/tests/CMakeLists.txt
@@ -8,6 +8,7 @@ add_executable(io-burner io-burner.cpp)
 target_link_libraries(io-burner PRIVATE Threads::Threads)
 
 add_executable(mem-burner mem-burner.cpp)
+target_compile_options(mem-burner PRIVATE "-fno-builtin")
 
 if (${CMAKE_VERSION} VERSION_GREATER "3.14.0" AND BUILD_GTESTS)`

Perhaps this is the easiest/laziest way. I can push it in this PR if you agree @graeme-a-stewart.

graeme-a-stewart:
I have a better fix in #212.

Getting CI tests to run against a clang compliation is a little more tricky - will be done in a later PR

PR #213, opened 2021-12-14
Clang CI option

### LINKED ISSUE #201: Cleanup CI

graeme-a-stewart:
For the c8-dev platform activate also a build with clang

We add an additional martix variable (compiler) that propages through to the environment, that triggers settng the C++ compiler to clang++. It would be then easy to add this compiler (or any other) for as many platforms as we think we need.

Closes #201

### DIFF
--- a/.github/scripts/build-test.sh
+++ b/.github/scripts/build-test.sh
@@ -9,14 +9,14 @@
 cd /tmp
 echo "Starting container run for platform $PLATFORM, compiler suite $COMPILER"
 if [ -z "$CXX" ]; then
-	if [ $COMPILER == "clang" ]; then
+	if [ "$COMPILER" == "clang" ]; then
 		CXX=$(type -p clang++)
 	fi
 	# gcc suite is the fall through
     CXX=$(type -p g++)
 fi
 if [ -z "$CC" ]; then
-	if [ $COMPILER == "clang" ]; then
+	if [ "$COMPILER" == "clang" ]; then
 		CXX=$(type -p clang)
 	fi
 	CC=$(type -p gcc)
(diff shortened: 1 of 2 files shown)


graeme-a-stewart:
HI @amete - this is ready to go now. We have a clang build on CentOS8 and it's easy to add other platforms if we felt the need. In parallel the containers were brought up to date and they all work find in the text matrix.

.github/scripts/build-test.sh
```diff
@@ -1,18 +1,35 @@
...
 # to cmake for particular platforms
 cd /tmp
+echo "Starting build and test for platform $PLATFORM, compiler suite $COMPILER"
 if [ -z "$CXX" ]; then
-    CXX=$(type -p g++)
+	if [ "$COMPILER" == "clang" ]; then
+		CXX=$(type -p clang++)
+	else
+		# gcc suite is the fall through
+    	CXX=$(type -p g++)
```
amete, line 16: Admittedly this is my OCD kicking in but can we indent this properly? 😄

amete (changes requested):
LGTM. I have only a minor cosmetic comment 👍

.github/scripts/build-test.sh
graeme-a-stewart: It seemed to be ok for me, but it was tabs - changed to space indentation (4 spaces), so then that should be more consistent

amete (approved):

PR #214, opened 2021-12-21
Add source release action

### LINKED ISSUE #209: Submodule missing from source tarball
The spdlog submodule is missing from the source tarball (https://github.com/HSF/prmon/archive/refs/tags/v3.0.0.tar.gz). Noticed this since i use the source tarball for an [AUR package of prmon](https://aur.archlinux.org/packages/prmon/). There are workarounds (i could directly clone the git or manually fetch the submodule or use the binary package), but maybe there is a way to include the submodule in the source tarball.

graeme-a-stewart:
As the GitHub source archives do not contain any
submodules they are not very useful for people
wishing to build a released
version of prmon. This action should build a source release
archive with resursion switched on at checkout, so that
spdlog will be included correctly.

Closes #209

### DIFF
--- a/.github/workflows/build-source-release.yml
+++ b/.github/workflows/build-source-release.yml
@@ -20,12 +20,12 @@ jobs:
       id: archive
       run: |
         VERSION=${{ github.event.release.tag_name }}
-        PKGNAME="prmon-source-$VERSION"
-        mkdir -p /tmp/$PKGNAME
-        mv * /tmp/$PKGNAME
-        mv /tmp/$PKGNAME .
-        TARBALL=$PKGNAME.tar.xz
-        tar cJf $TARBALL $PKGNAME
+        SRCNAME="prmon-$VERSION"
+        mkdir -p /tmp/$SRCNAME
+        mv * /tmp/$SRCNAME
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
Hi Serhan - in principle this should fix #209, but I am trying to think how to think if there's a way to test it without making half a dozen dummy releases.

amete (approved):
Looks good, thanks a lot @graeme-a-stewart. Let's see how this behaves w/ the next release.

graeme-a-stewart:
I think the unpacked directory name is going to be very slightly different, viz,. `prmon-v3.0.1`, but would that be a problem?

amete:
I think it's more of a convention than anything else. Therefore, I don't think it'll be a problem but we'll know for sure when we try to do a full deployment 😄

nikoladze:
Thanks a lot! I tried it in [my fork](https://github.com/nikoladze/prmon/releases/tag/vX.Y.Z2) and it seemed to work fine :slightly_smiling_face:

graeme-a-stewart:
Thanks a lot for checking @nikoladze !

PR #220, opened 2022-04-28
Support both external and internal dependencies

chuckatkins:
This adds support for both internal and external versions dependencies:

- `nlohmann_json` and `spdlog` are each supported now as as either externally found or internal submodules
  - submodules have been made also shallow
- Adds a reusable macro `_prmon_add_internal_external_dependency` used by both dependencies
- Adds a new configure option for each, `NLOHMANN_JSON` and `SPDLOG`:
  - `-DUSE_EXTERNAL_FOO={TRUE,FALSE,`**`AUTO`**`}`
    - `ON`, `TRUE`: Force an external version and fail if not found.
    - `OFF`, `FALSE`: Require the internal copy be used.
    - **`AUTO`**: Search for an external version and fall back to the internal copy if not found.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,8 @@
-[submodule "submodules/spdlog"]
-	path = submodules/spdlog
-	url = https://github.com/gabime/spdlog.git
 [submodule "submodules/nlohmann_json"]
+	shallow = true
 	path = submodules/nlohmann_json
 	url = https://github.com/nlohmann/json.git
+[submodule "submodules/spdlog"]
+	shallow = true
+	path = submodules/spdlog
+	url = https://github.com/gabime/spdlog.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "submodules/spdlog"]
 	path = submodules/spdlog
 	url = https://github.com/gabime/spdlog.git
+[submodule "submodules/nlohmann_json"]
+	path = submodules/nlohmann_json
+	url = https://github.com/nlohmann/json.git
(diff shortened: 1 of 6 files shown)


vvolkl:
Can confirm this works nicely both with submodules and external dependencies (which are preferred when building with spack). Thanks Chuck!

chuckatkins:
Updated to fix typos in README

chuckatkins:
Fixed typos in README

graeme-a-stewart:
Hi @chuckatkins - thanks very much for this. It seems to work very well for me, but I found an issue that our GoogleTest binaries had a missing CMake option, needed to find the spdlog headers. I have fixed this in 

https://github.com/HSF/prmon/tree/support-both-external-and-internal-dependencies

which has an extra commit. Do you want to pull from that branch or shall I pull that one directly?

graeme-a-stewart (approved):

graeme-a-stewart:
(And apologies for the late reply here - I was away from the office for 2 weeks)

graeme-a-stewart:
> > Hi @chuckatkins - thanks very much for this. It seems to work very well for me, but I found an issue that our GoogleTest binaries had a missing CMake option, needed to find the spdlog headers. I have fixed this in
> > https://github.com/HSF/prmon/tree/support-both-external-and-internal-dependencies
> > which has an extra commit. Do you want to pull from that branch or shall I pull that one directly?
> 
> I see the issue. I'll pull it into the branch and fix ci later this evening. I'm surprised it didn't come up when I was testing it. If it makes sense to do so I can also add ci builds to cover the different internal/external dependency combinations.

If you have the chance to add the extra CI tests please do! Otherwise I will create a follow-up issue and we'll take care of it.

We obviously don't need to test this on all platforms, either Ubuntu 20.04 or CentOS8 would suffice. I think it also only needs to be a build test.

PR #226, opened 2022-10-31
Update citation name and (C) year

graeme-a-stewart:
The original citation file had an odd capitalisation - changed to be `prmon`, all lower case, as used in documentation.

Update the overall (C) year to 2022.


amete (approved):
Thanks a lot @graeme-a-stewart, LGTM!

PR #228, opened 2023-05-17
Add long option to pstree to prevent truncating

miguelgila:
Hi,

I've noticed when there are a lot of levels of nested children processes/threads, the output of ```pstree -A -p %ld | tr \\- \\\\n``` doesn't always include the pids of all the children processes, as the output of `pstree` is truncated to 132 characters (the default). A workaround this limitation is to include the flag `-l` in the call to `pstree`, which avoids truncating altogether.

Alternatively, the default column where it starts truncating could be increased, but I think this workaround is cleaner.

What do you think?

Thanks,
Miguel

### DIFF
--- a/package/src/prmonutils.cpp
+++ b/package/src/prmonutils.cpp
@@ -38,7 +38,7 @@ std::vector<pid_t> pstree_pids(const pid_t mother_pid) {
   // method
   std::vector<pid_t> cpids;
   char smaps_buffer[64];
-  snprintf(smaps_buffer, 64, "pstree -A -p %ld | tr \\- \\\\n",
+  snprintf(smaps_buffer, 64, "pstree -l -A -p %ld | tr \\- \\\\n",
            (long)mother_pid);
   FILE *pipe = popen(smaps_buffer, "r");
   if (pipe == 0) return cpids;


graeme-a-stewart (approved):

graeme-a-stewart:
Hello @miguelgila - thanks very much for the patch. It looks fine to me.

OTOH, this older `pstree` method for finding all of the child PIDs is quite deprecated. Even the venerable CentOS7 kernel supports the task children attributes in `/proc` (see https://github.com/HSF/prmon/blob/46fbc5b2e8282974239558c7e24a269a1a4dfe66/package/src/prmonutils.cpp#L22).

Which system were you using that you came across this issue?

I don't mind accepting the patch, but likely we shall remove this support for very old kernels at some point.

graeme-a-stewart:
P.S. Apologies for sitting on this for so long. The email/alert from GitHub was overlooked and I just didn't see this for ages.

miguelgila:
Hi @graeme-a-stewart, thanks for accepting the patch. We are running this on a version of SLE 15 SP3 where I can't find the task children attributes. I can't say if it's because the upstream kernel doesn't have that capability, or because ours (provided by HPE) is modified and doesn't.

graeme-a-stewart:
Thanks @miguelgila - that's very interesting. So it's not to do with the age of the kernel, but also on some feature set that is exposed to user space. Then we won't deprecate this code ever, but we keep it for just such runtime environments.

PR #230, opened 2023-06-23
Cmake warning

### LINKED ISSUE #229: Need to update CMake policy
Our cmake step triggers now the following warning:

```
CMake Warning (dev) at /home/graemes/.local/cmake-3.26.4-linux-x86_64/share/cmake-3.26/Modules/FetchContent.cmake:1282 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
Call Stack (most recent call first):
  package/tests/CMakeLists.txt:20 (FetchContent_Declare)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

This should be fixed by adding the `DOWNLOAD_EXTRACT_TIMESTAMP OFF` when external packages are downloaded.

graeme-a-stewart:
Update the extraction policy for the GTest externals - this is because of recent CMake changes in 3.24 (see <https://cmake.org/cmake/help/latest/policy/CMP0135.html#policy:CMP0135>).

I also took the opportunity to update our submodules to the latest releases:
- nlohmman-json: 3.11.2
- spdlog: 1.11.0

N.B. I experimented with various options re. CMake policies, but the long and short of it is that adapting to this change means we need to have a minimum CMake version of 3.24, which is pretty new (August 2022). @amete - what's being used in ATLAS these days?

Probably it's also time for a minor release.

Closes #229

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,9 @@
 # - Define the minimum CMake version
 # HSF recommends 3.3 to support C/C++ compile features for C/C++11 across all
 # platforms
-cmake_minimum_required(VERSION 3.3)
+# Updated to 3.24 to ensure correct handling of extracting GTest from tarball
+# (See CMake policy CMP0135)
+cmake_minimum_required(VERSION 3.24)
 # - Call project() to setup system
 # From CMake 3, we can set the project version easily in one go
 project(prmon VERSION 3.0.2)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,15 @@
 # - Define the minimum CMake version
 # HSF recommends 3.3 to support C/C++ compile features for C/C++11 across all
 # platforms
-# Updated to 3.24 to ensure correct handling of extracting GTest from tarball
-# (See CMake policy CMP0135)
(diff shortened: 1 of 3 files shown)


graeme-a-stewart:
Oh err missus.... we need to update to this newer CMake on our CI platforms!

graeme-a-stewart:
Oh, I found out how to do this and stay compatible with older makes:

```cmake
if (POLICY CMP0135)
  cmake_policy(SET CMP0135 NEW)
endif (POLICY CMP0135)
```

Panic over! The new version fixes things for new CMakes and doesn't break things for older versions.

amete (approved):

PR #232, opened 2023-07-26
Support multiple plots

Cossack42:
### Multiple Plots

This allows the user to plot multiple prmon files on the same graph. 

As part of this we have divided the main function into separate sub-functions.

### Results

When graphing one set of data, the syntax of the input remains the same. The graphs on the original post are reproducible using the updated program.

Inputting multiple sets of data (by separating the filenames using commas) allows the plotting of multiple data sets on the same set of axes. For multiple data sets, dotted and dashed lines are used to distinguish each data set.

- CPU-time as a function of wall-time:

`prmon_plot.py --input prmon.txt,prmon.txt.1 --xvar wtime --yvar vmem,pss,rss,swap --yunit GB `

test5

- Rate of change in memory usage as a function of wall-time:

test6

Inputting up to 7  data sets is supported, although the legend does increase significantly in size and the data sets are more difficult to distinguish on the graph: it may be preferable to graph two sets of data at a time. 

-Memory as a function of wall-time using 3 data sets:

test10

### Issues

Plotting a stacked graph using multiple data sets is not supported (and will print an error) due to its lesser readability, although plotting a stacked graph using one set of data (stacking multiple variables) is still supported.

Committed by RAL Work Experience students Ismail Ryabchuk and Will Haywood under the supervision of Stewart Martin-Haugh (@StewMH).

### DIFF
--- a/package/index.html
+++ b/package/index.html
@@ -1,0 +1,2 @@
+test file
+
--- a/package/index.html
+++ b/package/index.html
@@ -1,2 +1,0 @@
-test file
-
--- a/package/scripts/prmon_plot.py
+++ b/package/scripts/prmon_plot.py
@@ -148,11 +148,11 @@ def check_variables(data, var, ylist):
             print(f"ERROR:: Variable {carg} is not available in one of the data sets")
 
 
-# This function does something
+# This function creates the final data set of y-values
 
 def create_list(ylist, data, args, xmult, ymult, xlabel):
     ydlist = []
-    for carg in range(len(ylist)):
+    for carg in ylist:
         if args.diff:
             num = np.array(data[carg].diff()) * ymult
             denom = np.array(data[xlabel].diff()) * xmult
@@ -162,7 +162,6 @@ def create_list(ylist, data, args, xmult, ymult, xlabel):
             ydlist.append(np.array(data[carg]) * ymult)
     return ydlist
 
-
 # Graph plotting functions
 
 def draw_stacked_graph(xdata, ydlist, ylist):
@@ -171,9 +170,12 @@ def draw_stacked_graph(xdata, ydlist, ylist):
         xdata, ydata, lw=2, labels=[LEGENDNAMES[val] for val in ylist], alpha=0.6
     )
 
-def draw_line_graph(xdata, ydlist, ylist):
+def draw_line_graph(xdata, ydlist, ylist, style):
+    colours = plt.rcParams['axes.prop_cycle'].by_key()['color']     # This is a list of the matplotlib default colours
+    count = 0
     for cidx, cdata in enumerate(ydlist):
-        plt.plot(xdata, cdata, lw=2, label=LEGENDNAMES[ylist[cidx]])
+        plt.plot(xdata, cdata, lw=2, label=LEGENDNAMES[ylist[cidx]], color=colours[count], linestyle=style)
+        count += 1
 
 
 # Main function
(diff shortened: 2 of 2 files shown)


package/index.html
```diff
@@ -0,0 +1,2 @@
+test file
```
amete, line 1: I'm not sure what this file is for. Can you drop it from your PR?

package/scripts/prmon_plot.py
```diff
@@ -193,24 +252,20 @@ def main():
...
-    data = pd.read_csv(args.input, sep="\t")
-    data["Time"] = pd.to_datetime(data["Time"], unit="s")
+    inputs = args.input.split(",")
+    ylist = args.yvar.split(",")
+    data = []
 
-    # Check the variables are in data
-    if args.xvar not in list(data):
-        print(f"ERROR:: Variable {args.xvar} is not available in data")
+    if len(inputs) > 7:
```
amete, line 259: Why `7`? I think we can drop this limitation and let the user decide if the final plot looks overcrowded or not.

package/scripts/prmon_plot.py
```diff
@@ -257,24 +312,29 @@ def main():
...
+            for i in range(len(xdata)):
+                draw_stacked_graph(xdata[i], ydlist[i], ylist)
+        else:
+            print("ERROR:: Stacked graphs are not supported for more than one data set")
+            sys.exit(-1)
     else:
-        for cidx, cdata in enumerate(ydlist):
-            plt.plot(xdata, cdata, lw=2, label=LEGENDNAMES[ylist[cidx]])
+        for i in range(len(xdata)):
+            draw_line_graph(xdata[i], ydlist[i], ylist, line_styles[i % 7], inputs, i)
```
amete, line 335: If we drop the magical `7` then this'll need to be updated.

amete (changes requested):
Hi @Cossack42,

Thank you very much for your contribution. This is certainly a good functionality to have and the sample plots you attached look great. I have a few minor comments. Afterwards, this PR should be good to go.

Best,
Serhan

package/index.html
Cossack42: Yeah no problem.

package/scripts/prmon_plot.py
Cossack42: The reason the MOD 7 is used is because the list of line styles is only of length 7. Thus if we get more than 7 types of line we will get a index out of range.

package/scripts/prmon_plot.py
Cossack42: We have removed the limit, although the reason for 7 was that there was only 7 line types.

package/index.html
amete: If you can delete this file and add/commit/push your changes to your branch then I'll approve the PR and merge it.

Cossack42: Alright.

package/scripts/prmon_plot.py
```diff
@@ -257,24 +306,29 @@ def main():
...
+            for i in range(len(xdata)):
+                draw_stacked_graph(xdata[i], ydlist[i], ylist)
+        else:
+            print("ERROR:: Stacked graphs are not supported for more than one data set")
+            sys.exit(-1)
     else:
-        for cidx, cdata in enumerate(ydlist):
-            plt.plot(xdata, cdata, lw=2, label=LEGENDNAMES[ylist[cidx]])
+        for i in range(len(xdata)):
+            draw_line_graph(xdata[i], ydlist[i], ylist, line_styles[i%7], inputs, i)
```
graeme-a-stewart, line 329: This line fails a [flake8 test](https://www.flake8rules.com/rules/E228.html). It should be:

```py
draw_line_graph(xdata[i], ydlist[i], ylist, line_styles[i % 7], inputs, i)
```

We should fix that too.

graeme-a-stewart (changes requested):
Thanks so much for this @Cossack42, it's a nice feature to have. There's one small formatting bug to fix, then it's good to go from my side.

package/scripts/prmon_plot.py
Cossack42: Fixed!

graeme-a-stewart (approved):

package/scripts/prmon_plot.py
graeme-a-stewart: However, you're using `7` as a magic number - what if a future release of matplotlib has more or less line styles?

At L318 you have:

```py
    line_styles = list(mpl.lines.lineStyles.keys())
```

Which suggests that the correct code would be `line_styles[i % len(line_styles)]`, right?

Cossack42: Yes, that's a good point. Making these changes will ensure compatibility with future updates. I have made the necessary changes!

graeme-a-stewart: Thank you, looks good now!

Cossack42: Ah no worries. I just reformatted it with black.

amete (approved):

amete:
Thanks a lot @Cossack42, @graeme-a-stewart. All looks good now.

Cossack42:
No thank you! I enjoyed working on this.

graeme-a-stewart:
Let me add again my thanks @Cossack42. These are great contributions to receive. 🙏

Cossack42:
Anytime!

PR #234, opened 2023-08-03
Add support for fast memory monitoring using smaps_rollup

### LINKED ISSUE #219: Add support for smaps_rollup
Hi @graeme-a-stewart,

Arguably one of the most expensive operations we do is traversing through `smaps` to get an accurate measurement of `pss`. It seems as if the "newer" kernel version have a new file, called `/proc/[pid]/smaps_rollup`, that provides pre-summed memory information for a given process. It looks to be a very convenient way to simply read the `pss` measurement rather than reading/summing everything by-hand. We can probably add support for it some time soon (i.e. use it if it's available). What do you think?

Best,
Serhan

amete:
This PR adds the so-called fast memory monitoring option, which uses the `smaps_rollup` file instead of the standard `smaps` file, as discussed in #219.

There are some limitation though. Most notably the `Size` metric is missing in the pre-summed file, where we get the virtual memory measurements, and the fact that `smaps_rollup` is not supported by some older kernels, e.g., `cc7`. Therefore, we probably need to do a little bit of testing before using this new mode for anything serious.

Closes #219

### DIFF
--- a/package/src/memmon.cpp
+++ b/package/src/memmon.cpp
@@ -177,4 +177,4 @@ void const memmon::do_fastmon() {
       it++;
     }
   }
-};
+}
--- a/package/src/memmon.cpp
+++ b/package/src/memmon.cpp
@@ -18,7 +18,7 @@
 
 // Constructor; uses RAII pattern to be valid
 // after construction
-memmon::memmon() {
+memmon::memmon() : input_filename{"smaps"} {
   log_init(MONITOR_NAME);
 #undef MONITOR_NAME
   for (const auto& param : params) {
@@ -35,7 +35,8 @@ void memmon::update_stats(const std::vector<pid_t>& pids,
   std::string key_str{}, value_str{};
   for (const auto pid : pids) {
     std::stringstream smaps_fname{};
-    smaps_fname << read_path << "/proc/" << pid << "/smaps" << std::ends;
+    smaps_fname << read_path << "/proc/" << pid << "/" << input_filename.c_str()
+                << std::ends;
     std::ifstream smap_stat{smaps_fname.str()};
     while (smap_stat) {
       // Read off the potentially interesting "key: value", then discard
--- a/package/src/memmon.cpp
+++ b/package/src/memmon.cpp
@@ -170,7 +170,7 @@ void const memmon::do_fastmon() {
   for (auto it = mem_stats.cbegin(); it != mem_stats.cend();) {
     // Delete unavailable metrics
     if (available_metrics.count(it->first) == 0) {
-      spdlog::warn("Metric " + it->first +
+      spdlog::info("Metric " + it->first +
                    " is not available in fast monitoring");
       it = mem_stats.erase(it);
     } else {
--- a/package/src/memmon.cpp
+++ b/package/src/memmon.cpp
@@ -10,6 +10,7 @@
 #include <iostream>
 #include <limits>
 #include <regex>
+#include <set>
 #include <sstream>
 
 #include "utils.h"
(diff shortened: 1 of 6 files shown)


package/src/prmon.cpp
```diff
@@ -263,10 +278,11 @@ int main(int argc, char* argv[]) {
       {"netdev", required_argument, NULL, 'n'},
       {"help", no_argument, NULL, 'h'},
       {"level", required_argument, NULL, 'l'},
+      {"fast-memmon", no_argument, NULL, 'm'},
       {0, 0, 0, 0}};
 
   int c;
-  while ((c = getopt_long(argc, argv, "-p:f:j:i:d:sun:h:l:", long_options,
+  while ((c = getopt_long(argc, argv, "-p:f:j:i:d:sun:h:l:m:", long_options,
```
graeme-a-stewart, line 285: Should this be `"-p:f:j:i:d:sun:h:l:m"`? The `-m` option doesn't take an argument, it's a toggle.

graeme-a-stewart (changes requested):
There's one minor problem with the `get_longopts` that's trivial.

My larger concern is that when you switch this option on the reported `vmem` is `0`. That's just wrong and I don't like using `0` as a magic value that means *sorry, we couldn't measure this*.

My proposal is that we'd rather drop that value from the reported statistics, which would be less confusing. Plus, that definitely needs documentation!

package/src/prmon.cpp
amete: Yep, done now 👍

amete:
Wow, I didn't realize this slipped through the cracks for so long @graeme-a-stewart 😄 Thanks a lot for the review. I tweaked the implementation in that if a particular metric is now available, we print a warning message and drop it from the list of monitored metrics. I also fixed the type in the options. How does this look from a functional point of view? If it looks OK, I can put together a few words in the README for the new functionality and we can probably push this through. Please let me know, many thanks.

package/src/memmon.cpp
```diff
@@ -135,3 +137,44 @@ void const memmon::get_unit_info(nlohmann::json& unit_json) {
...
+      }
+    }
+  }
+
+  // Delete unavailable metrics from the monitored stats
+  // In C++17/20 there are more elegant ways to do this
+  for (auto it = mem_stats.cbegin(); it != mem_stats.cend();) {
+    // Delete unavailable metrics
+    if (available_metrics.count(it->first) == 0) {
+      spdlog::warn("Metric " + it->first +
```
graeme-a-stewart, line 173: Shouldn't this be "info" instead of "warn"? Warn seems a bit hard when the user actually asked for it (`--fast-memmon`)

graeme-a-stewart (changes requested):
Hi @amete - this is looking good now. I only have one code suggestion (downgrade the warning to info). And the README will need to be updated.

package/src/memmon.cpp
amete: The logic behind the warning was that although the user asked for fast monitoring it might be unbeknownst to them that certain metrics might be missing in that configuration (although we can document this in the README etc.). Having said that, info is also OK w/ me.

graeme-a-stewart: I am just wary of forcing warnings that will just be ignored (cf. ATLAS reconstruction jobs in Run 1!)

amete: Believe it or not we had the same discussion just two days ago 🤣  History repeats itself. Let's print this as an info message and I'll add some text to README.

amete:
Thanks a lot @graeme-a-stewart. I just pushed a new version. Can you let me know how this looks, please?

graeme-a-stewart (approved):
Looks good - thanks @amete

PR #236, opened 2023-12-01
prmon_plot: don't use square brackets around units

olifre:
Using square brackets around units hurts the heart of the physicist, as square brackets must only be used around quantities to denote their units (i.e. [F] = N).  The correct notation recommended by SI, ISO and DIN standards is to divide by the unit instead.

see also https://academia.stackexchange.com/a/28918 for some more references on this matter

### DIFF
--- a/package/scripts/prmon_plot.py
+++ b/package/scripts/prmon_plot.py
@@ -348,8 +348,8 @@ def main():
         fyunit = args.yunit
 
     plt.title("Plot of {} vs {}".format(fxlabel, fylabel), y=1.05)
-    plt.xlabel((fxlabel + " [" + fxunit + "]") if fxunit != "1" else fxlabel)
-    plt.ylabel((fylabel + " [" + fyunit + "]") if fyunit != "1" else fylabel)
+    plt.xlabel((fxlabel + " / " + fxunit) if fxunit != "1" else fxlabel)
+    plt.ylabel((fylabel + " / " + fyunit) if fyunit != "1" else fylabel)
     plt.tight_layout()
     fig.savefig(output)
 


amete:
Thanks for your contribution @olifre. This choice stems from the [ATLAS Experiment's Style Guide](https://cds.cern.ch/record/1110290), which reads:

```
All axes must be labelled, including units. Units should be given in square brackets,
e.g. ‘Energy [GeV]’.
```

under `3.3.3 General RULES and RECOMMENDATIONS for graphs and plots`. Admittedly, it's an "ATLAS choice" and given the informal nature of the plots we produce here I'd be OK with anything as long as the message is clear. Let me ping @graeme-a-stewart to see if he has any preference either way.

olifre:
@amete Thanks for the feedback! 

My reasoning is that at our University we are asked to teach students to correctly indicate units (since there is an ISO and DIN standard which explicitly say that square brackets must not be used, and an [SI recommendation](https://www.bipm.org/en/publications/si-brochure) to divide by the unit). I also know of cases in industry when official reports were sent back by an industry partner since the units were wrongly given in square brackets, and work had to be re-done. 

As the scientific career (for many students) is only a part of their lives preparing them for a job in industry, it seems best to adhere to official standards whenever possible to best prepare them for their life out there, so they don't run into a similar kind of mess in their potential later career in industry. 

Of course, giving units in a non-ISO non-SI non-DIN way is never "punished" in science — but since `prmon` is a tool which is of very general use, I'd love to see it be close to those standards to ease use of the produced plots outside of an experiment-specific context.

graeme-a-stewart:
Hi both - sorry for not contributing here until now, this got a bit lost pre-Christmas holidays.

FMyI I had a quick look at some recent LHC papers and CMS follow the same standard as ATLAS. ALICE and LHCb seem to be a bit inconsistent, sometimes they use ()s, sometimes []s. APS and Nature journals very much favour ()s for the units.

That said, I'm mildly in favour of following the SI guidelines for prmon, even if they are not very much used in HEP - we're measuring `s`, `GB`, etc. here so the meaning if pretty clear.

graeme-a-stewart (approved):

PR #238, opened 2024-04-29
Update output parsing for nvidiamon

elmsheus:
As discussed in the ATLAS ticket https://its.cern.ch/jira/browse/ATEAM-985, this PR updates the output parsing of "nvidia-smi pmon -s um -c 1" and add the 3 missing variables.

Tagging @amete and @graeme-a-stewart for info and feedback.

PR #239, opened 2024-05-03
Nvidia monitor fixes

graeme-a-stewart:
As noticed by @elmsheus, the output fields from the `nvidia-smi pmon` have changed, which requires an update to the parsing from prmon.

This PR builds on #238 and fixes the reference files used by prmon for "stand alone" testing.

To help with testing GPU parse code, a small burner script, using `pycuda` is added, `gpu-burner.py`.

Closes #238


krasznaa:
Hi Graeme, Serhan,

For the longer term, I wonder if we could learn from what [Syllo/nvtop](https://github.com/Syllo/nvtop) is doing. :thinking: I have to admit that I also only learned about that project in the last few days. But it is **super nice**.

image

It seems to extract information from NVIDIA (and AMD, Intel) GPUs in C(\+\+). And it's a lot more information than what `nvidia-smi pmon` is giving out. (Including, I think very interestingly, the GPU's estimate for how much power it is using at any given moment.)

This may be a very good project for a GSOC, or some other type of student to investigate. :wink:

Cheers,
         Attila

Issue #276, opened 2025-12-26
feat: Add cgroup v2 support for accurate container resource monitoring


DhanashreePetare:
### **Problem**
Modern Linux distributions and containers (Docker/Kubernetes) use cgroup v2. Current prmon monitoring has reduced accuracy in containerized environments and cannot detect container resource limits.

### **Proposed Solution**
Add new `cgroupmon` monitor component that:
- Auto-detects and reads from cgroup v1/v2 (`/sys/fs/cgroup/`)
- Provides accurate CPU, memory, I/O stats for containers
- Reports container limits and quotas
- Maintains backward compatibility with non-container environments

### **Expected Deliverables**
- [ ] `cgroupmon.cpp/h` implementing `Imonitor` interface
- [ ] Support for both cgroup v1 and v2
- [ ] Auto-detection with fallback to `/proc`
- [ ] Unit tests with precooked sources
- [ ] Integration tests in Docker/Kubernetes
- [ ] Documentation updates (README, ADDING_MONITORS.md)

### **Benefits**
- Accurate monitoring in modern container platforms
- Lightweight implementation (no new dependencies)
- Suitable for HEP workflows in Kubernetes

PR #277, opened 2025-12-26
Feat/cgroup v2 support (#276)

DhanashreePetare:
### Overview
Adds comprehensive cgroup (container) resource monitoring to prmon with automatic detection of cgroup v1, v2, and hybrid environments.

### Changes
- **New monitor**: `cgroupmon` with 17 metrics across CPU, memory, I/O, and process counts
- **Auto-detection**: Detects cgroup version at runtime and adapts accordingly
- **Parser**: Handles `cpu.stat`, `memory.stat`, `io.stat` for both v1 and v2
- **Tests**: Python integration test + GTest skeleton with precooked fixtures
- **Docs**: README section, ADDING_MONITORS example, CONTRIBUTING Windows guide
- **Disable flag**: `--disable cgroupmon` if needed

### Benefits
- Better container/Kubernetes workload monitoring
- More accurate I/O stats in containerized environments
- Backward compatible with existing prmon deployments
- Graceful degradation if cgroups unavailable

### Testing
Python linting passes. CI will verify build and tests in Linux environments.


graeme-a-stewart (changes requested):
See comment above for restructuring changes.

graeme-a-stewart:
Hello @DhanashreePetare 

Is this an issue you would still like to work on? Please let us know and we can try to assign to another person if you are not able to continue.

Thanks.

DhanashreePetare:
sorry @graeme-a-stewart, very sorry for my late reply. Yes I will change the workflow according to the above provided changes, I will get back very soon with the changes.

doc/CONTRIBUTING.md
```diff
@@ -65,6 +65,46 @@ issue to first discuss with the maintainers.
...
+## Windows Development Notes
+
+The scope of `prmon` is Linux systems (access to `/proc` and cgroups is
+required). If you develop on Windows, we recommend building and running tests
+inside a Linux environment:
+
+- Use WSL2 (Ubuntu) or Docker Desktop to provide a Linux runtime.
+- Run all builds and tests inside WSL2 or a Linux container.
+
+For formatting on Windows before opening a pull request:
```
graeme-a-stewart, line 85: Windows is not a supported environment for prmon, so while I appreciate your instructions, it's just adding noise. If one is using Windows or OS X then a container / WSL2 environment needs to be used and then you have your Linux for build, tests, formatting, etc.

doc/CONTRIBUTING.md
```diff
@@ -65,6 +65,46 @@ issue to first discuss with the maintainers.
...
+Get-ChildItem src, package/src, tests -Recurse -Include *.cpp,*.h |
+  ForEach-Object { clang-format -i $_.FullName }
+```
+
+Alternatively, when using CMake, you can invoke the `clang-format` target (if
+available):
+
+```powershell
+cmake -S . -B build
+cmake --build build --target clang-format
```
graeme-a-stewart, line 110: This is already documented (L59).

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
+// Copyright (C) 2018-2025 CERN
```
graeme-a-stewart, line 1: This is a new file, so it should be (C) for the dates the file existed, viz, 2025-2026 in this case.

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+  cgroup_file << "/proc/" << pid << "/cgroup";
+  
+  std::ifstream cgroup_stream(cgroup_file.str());
+  if (!cgroup_stream.is_open()) {
+    return "";
+  }
+  
+  std::string line;
+  // For cgroup v2, look for line starting with "0::"
+  // For cgroup v2, look for line starting with "0::"
```
graeme-a-stewart, line 75: Duplicated comment.

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+    }
+  }
+  
+  return "/";  // Root cgroup
+}
+
+// Find the cgroup filesystem mount point
+std::string cgroupmon::find_cgroup_mount_point() {
+  return "/sys/fs/cgroup";
+}
```
graeme-a-stewart, line 88: This is now OTT for returning the same string every time (and you're not "detecting" anything). Instead just define a `const std::string`.

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+  log_init(MONITOR_NAME);
+#undef MONITOR_NAME
+
+  // Initialize monitored values
+  for (const auto& param : params) {
+    cgroup_stats.emplace(param.get_name(), prmon::monitored_value(param));
+  }
+
+  // Detect cgroup version
+  cgroup_version = detect_cgroup_version();
```
graeme-a-stewart, line 33: This now seems odd - we only support one version (v2) and we just detect if this seems to be supported. `CgroupVersion::NONE` isn't really a version at all.

Can you change the function naming and logic here to then be a bit clearer? e.g., you can have `detect_cgroup_v2()` returning a boolean instead. I think this will work and it will be clearer.

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+  
+  if (cgroup_version == CgroupVersion::NONE) {
+    info("No cgroup support detected, cgroupmon will be inactive");
+    return;
+  }
+
+  // Find cgroup mount point
+  cgroup_mount_point = find_cgroup_mount_point();
```
graeme-a-stewart, line 41:
```cpp
const std::string cgroup_mount_point = "/sys/fs/cgroup";
```

would be clearer and avoid a useless function.

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+  if (cgroup_mount_point.empty()) {
```
graeme-a-stewart, line 42: Right now this does nothing at all, as `cgroup_mount_point` is always defined. A test that the mount point exists has already been done in `detect_cgroup_version()`.

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+// Update statistics from cgroup
+void cgroupmon::update_stats(const std::vector<pid_t>& pids,
+                             const std::string read_path) {
+  if (!valid || pids.empty()) {
+    return;
+  }
+
+  // Use first PID to determine cgroup path
+  // (all processes in the tree should be in the same cgroup)
+  std::string base_path = read_path.empty() ? "" : read_path;
```
graeme-a-stewart, line 99: `base_path` serves no purpose - it's always the same as `read_path`. And the ternary operation is pointless.

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+  // CPU statistics
+  parse_cpu_stat_v2(cgroup_path + "/cpu.stat", stats);
+  
+  // Memory statistics
+  parse_memory_stat_v2(cgroup_path + "/memory.stat", stats);
+  
+  // Memory current and max
+  stats["cgroup_mem_current"] = read_single_value(cgroup_path + "/memory.current") / 1024;
+  unsigned long long mem_max = read_single_value(cgroup_path + "/memory.max");
+  if (mem_max != ULLONG_MAX) {
```
graeme-a-stewart, line 125: First, better to use the modern C++ limit here, viz. `std::numeric_limits<unsigned long long>::max()`.

Second, what is the purpose of this test?

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+  
+  // CPU statistics
+  parse_cpu_stat_v2(cgroup_path + "/cpu.stat", stats);
+  
+  // Memory statistics
+  parse_memory_stat_v2(cgroup_path + "/memory.stat", stats);
+  
+  // Memory current and max
+  stats["cgroup_mem_current"] = read_single_value(cgroup_path + "/memory.current") / 1024;
+  unsigned long long mem_max = read_single_value(cgroup_path + "/memory.max");
```
graeme-a-stewart, line 124: What happens when `memory.max` has no limit and contains the entry `max`? As far as I can see you would here return `0`, correct?

On reflection I think it is probably the most useful thing to map a `max` value to zero, but this needs to be well documented (it can happen for some other cgroupsv2 values as well).

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+    while (std::getline(file, line)) {
+      if (!line.empty()) {
+        count++;
+      }
+    }
+    return count;
+  }
+
+  // For single-value files, read the first number
+  unsigned long long value;
```
graeme-a-stewart, line 242: I think the code here would be better if you mapped the one expected string, `max`, to the value `0` explicitly.

At the moment this is happening silently and implicitly, which is not good.

package/src/cgroupmon.cpp
```diff
@@ -0,0 +1,291 @@
...
+// Return parameter list
+prmon::parameter_list const cgroupmon::get_parameter_list() { return params; }
+
+// Collect hardware information
+void const cgroupmon::get_hardware_info(nlohmann::json& hw_json) {
+  if (!valid) {
+    return;
+  }
+  
+  hw_json["cgroup"]["version"] = "v2";
```
graeme-a-stewart, line 284: This is not useful anymore as cgroups implies v2, and the mount point is fixed. Please remove.

package/src/cgroupmon.h
```diff
@@ -0,0 +1,95 @@
...
+#include <string>
+#include <vector>
+
+#include "Imonitor.h"
+#include "MessageBase.h"
+#include "parameter.h"
+#include "registry.h"
+
+// Cgroup version enumeration
+enum class CgroupVersion { NONE, V2 };
```
graeme-a-stewart, line 20: I think this is no longer needed - see comments above w.r.t. detecting cgroupsv2. An enum is overdesign.

package/src/cgroupmon.h
```diff
@@ -0,0 +1,95 @@
...
+      {"cgroup_io_write", "B", "B/s"},
+      // Process/thread counts from cgroup
+      {"cgroup_nprocs", "1", "1"},
+      {"cgroup_nthreads", "1", "1"}};
+
+  // Dynamic monitoring container
+  prmon::monitored_list cgroup_stats;
+
+  // Cgroup detection state
+  CgroupVersion cgroup_version;
```
graeme-a-stewart, line 49: Not needed.

package/src/cgroupmon.h
```diff
@@ -0,0 +1,95 @@
...
+  bool valid;
+  std::string cgroup_path;
+  std::string cgroup_mount_point;
```
graeme-a-stewart, line 52: This could now just be a `const std::string`.

CGROUP_IMPLEMENTATION.md
```diff
@@ -0,0 +1,239 @@
+# Cgroup v2 Support Implementation - Summary
```
graeme-a-stewart, line 1: We really don't need this file. The historical record of the addition of the cgroups monitoring is in the pull request.

graeme-a-stewart (changes requested):
Hello @DhanashreePetare - thanks a lot for the changes. There are still a few things to improve on with some further simplifications that will make the code even leaner.

graeme-a-stewart:
FYI, I just fixed the `CMakeLists.txt` conflict for you

graeme-a-stewart:
There is also a test failure, can you investigate it?

```sh
25/28 Testing: testUnits
25/28 Test: testUnits
Command: "/usr/bin/python3" "test_cpu.py" "--units" "--time" "3" "--slack" "0" "--interval" "1"
Directory: /tmp/build/package/tests
"testUnits" start time: Mar 24 17:33 UTC
Output:
----------------------------------------------------------
Will run for 3s using 1 process(es) and 1 thread(s)
[2026-03-24 17:33:21.529] [nvidiamon] [warning] Failed to load libnvidia-ml.so: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
[2026-03-24 17:33:21.529] [nvidiamon] [warning] Both NVML and nvidia-smi failed. NVIDIA monitoring disabled.
F
======================================================================
FAIL: test_run_test_with_params (__main__.setup_configurable_test.<locals>.ConfigurableProcessMonitor)
Actual test runner
----------------------------------------------------------------------
  File "/tmp/build/package/tests/test_cpu.py", line 99, in test_run_test_with_params
    self.assertEqual(
AssertionError: 3 != 0 : Wrong number of unit values for 'Avg' - missing parameters are {'cgroup_cpu_periods', 'cgroup_cpu_throttled', 'cgroup_mem_max'}

----------------------------------------------------------------------
Ran 1 test in 3.075s
```

N.B. You can ignore the `nvidiamon` warning - that needs to be lowered in severity

graeme-a-stewart:
Hello @DhanashreePetare - do you have interest in continuing this PR. We still have a few outstanding issues.

PR #297, opened 2026-02-14
Direct parsing for Nvidia devices from NVML

### LINKED ISSUE #242: Convert Nvidia monitoring to use C API
Instead of parsing `nvidia-smi` prmon should use the more stable API calls like [this](https://github.com/Syllo/nvtop/tree/master/src).

GPU monitoring components will need to become a build option.

EyupKeremBas7:
Fixes #242.

As discussed in the issue by @graeme-a-stewart, this PR replaces the `nvidia-smi` subprocess parsing with direct calls to the NVML C API.

**Key Changes & Architecture:**

* **Build Option:** As requested, GPU monitoring is now an opt-in build feature. Added the `ENABLE_NVIDIA_GPU` CMake flag (defaults to `OFF` to prevent build failures on non-GPU systems).

* **Code Style & RAII:** I strictly maintained the original RAII pattern of the `nvidiamon` class (initializing NVML in the constructor and shutting it down in the destructor). The logging style and overall class structure were kept as close to the original as possible (for example, retaining the `read_path` parameter in the `update_stats` signature to preserve the existing interface, even though it is no longer in use).

* **Note on Loops:** You might notice the use of classic index-based `for` loops instead of range-based C++ iterators in `update_stats`. This was necessary to properly handle the C-style arrays and pointers required by the NVML functions.

* **Safe Fallback:** If the binary is compiled with NVML enabled but runs on a machine without NVIDIA drivers/GPUs, it simply logs a warning and disables the GPU monitoring without crashing.

**Testing:**

To ensure zero regressions and verify the exact same output format for downstream tools (like `prmon_plot.py`), I created a custom automated test suite. I ran a matrix of 54 tests comparing the legacy `nvidia-smi` parser vs. the new NVML implementation. Scenarios included CPU-only, GPU-heavy, and Parent/Child forking across three environments:

1. Google Colab (Tesla T4)

2. Local Docker (Ubuntu 22.04) 

3. Local Bare Metal (4050 RTX GPU) 

* The test scripts used are available here: https://github.com/EyupKeremBas7/prmon-test

* All raw test data and benchmark logs in this Drive folder: https://drive.google.com/file/d/1QwMiAY_u8rLbibOdik56bOhTKtAwWvpO/view?usp=sharing

**Performance Validation Note:** The plot below illustrates the resource tracking of the dummy_gpu process from the prmon-test repository. This specific benchmark was executed on a Google Colab T4 instance for 30 seconds with a 1-second sampling interval to verify NVML parity with legacy outputs.

comparison_smi_vs_nvml_clean

---

**AI Assisted Contributions Disclosure:**

In accordance with the guidelines introduced in *Add section on AI contributions (#291)* (commit `d52e11d`), I confirm that I take full responsibility for all code and results submitted in this PR. For full transparency regarding non-routine tasks: the core C++ and CMake implementations in this PR were written entirely by me (using AI strictly for routine tasks like code review). However, to accelerate the extensive benchmarking process, the auxiliary test scripts in the `prmon-test` repository, as well as the code comments and this documentation, were generated with AI assistance.


graeme-a-stewart (changes requested):

graeme-a-stewart:
Thank you for the patch here @EyupKeremBas7. It's really great that you contribute to the project, 🙏.

However we have some issues needing to be addressed.

The way you find and use the CUDA library is very "old school" and we should update to a more modern approach, illustrated in the line-by-line comments above.

For my next comment, let me prefix by highlighting that for the main use of prmon in WLCG the experiments usually build the library themselves, deploying it to a shared filesystem called CVMFS (which is accessible worldwide). So there is *one build* and *many executions on different machines*. When prmon executes it needs to be able to *dynamically* decide if GPU monitoring should be enabled.

Currently, we would be in the following situation with this PR:

- Scenario 1 - build machine has no CUDA, no support added to prmon, no monitoring possible even on WLCG GPU nodes
- Scenario 2 - build machine has CUDA and we enable it, now prmon depends on `libnvidia-ml.so` and if it cannot be found the monitoring won't even run:

```sh
lxplus928 :: ~/build/prmon » ./package/prmon --help                                                                                     130 ↵
./package/prmon: error while loading shared libraries: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
```

1 are bad, and 2 is really terrible!

To address this we would need to rather be able to dynamically load the Nvidia monitoring code for prmon as a module (`dlopen()`), aborting any GPU monitoring if that fails. This will also probably impact on our ability to build statically - we'd never be able to support GPU monitoring in this case (I tried and I got a loader error, even though `libnvidia-ml.a` does exist).

So it's tricky....

I also worry about compatibility with different CUDA versions - I don't know how stable the NVML interfaces are.

A better approach is to enable the NVML monitoring as a *different monitoring plugin*, and that we keep the old style `nvidia-smi` external monitoring for a transition period, until we are confident that we addresses all of the deployment issues.

Please let @amete and me know if you would be able to work with us to address these issues.

graeme-a-stewart (changes requested):
To be restructured into a single monitor, supporting both nvml and nvidia-smi methods.

graeme-a-stewart (changes requested):
Thank you @EyupKeremBas7 - I think we are in pretty good shape now. I have added a few minor comments to address before we merge.

The point is taken that `nvidiamon.cpp` is quite a large source file now, but I believe the structure is clear and nothing is very difficult to understand. Conforming to the architectural design, that monitors are independent, puts us into a much better shape that we had when two monitoring methods that were intertwined.

graeme-a-stewart (approved):
Thank you so much @EyupKeremBas7! We are ready to go.

I had very tiny differences in the clang-format output that I got in `nvidiamon.cpp` - I just pushed my version from running:

```sh
clang-format -i --style=Google package/src/nvidiamon.*
```

That was with `clang-format version 20.1.8 (AlmaLinux OS Foundation 20.1.8-3.el9)`.

EyupKeremBas7:
Hi @graeme-a-stewart,

Thank you for the final formatting touch and for merging the PR! It’s been an incredible learning experience working on this implementation with you.

Since this is my first major open-source contribution to a project of this scale, I am planning to share this on my LinkedIn. I would love to tag you in the post to publicly thank you for your mentorship and guidance, if you are comfortable with that.

Thanks again for everything!

graeme-a-stewart:
I'm glad that you found the process useful and, sure, no problem to tag me in your post!

PR #307, opened 2026-03-03
Fix clang format

graeme-a-stewart:
Try to fix the clang-format CI action, which is not currently being run.

- Ensure that we only care about our source files (in `package`, ignored submodules)
- Run on pull request


graeme-a-stewart:
Annoyingly still seems to be broken @amete - any clues?

PR #311, opened 2026-05-07
feat: add nfds (number of file descriptors) monitoring to countmon

arch-yunus:
This PR introduces a new metric, nfds, to the countmon monitor. It tracks the total number of open file descriptors across the monitored process tree.

Key Changes:

Implemented prmon::count_fds utility to count entries in /proc/[pid]/fd.
Updated countmon to include nfds in the monitored statistics.
Aggregated the file descriptor count for all child processes during each monitoring interval.
Why this is useful: Monitoring the number of open file descriptors is crucial for identifying resource leaks and understanding the I/O footprint of complex, long-running processes in HEP workflows.

Testing: The implementation uses the standard /proc interface, consistent with existing monitors. It has been integrated into the countmon update loop which is already part of the core monitoring cycle.

### DIFF
--- a/package/src/countmon.cpp
+++ b/package/src/countmon.cpp
@@ -47,6 +47,7 @@ void countmon::update_stats(const std::vector<pid_t>& pids,
       count_stat_update["nprocs"] += 1L;
       count_stat_update["nthreads"] +=
           std::stol(stat_entries[prmon::num_threads]);
+      count_stat_update["nfds"] += prmon::count_fds(pid, read_path);
     }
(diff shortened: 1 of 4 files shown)


mohitt31:
Hello, I was looking to add nfds monitoring and  found out your PR.
The core logic is fine but its missing the integerations tests that CONTRIBUTING.md asks for
I've got an implementation with precooked test data and a test_count.py update ready - are you still working on this? If not I can open a separate PR with tests included.

PR #318, opened 2026-08-10
More robust parsing of GPU parameters

### LINKED ISSUE #317: Mitigate against "unusual" output from GPU hardware probing
GPU stats are obtained via an `nvidia-smi` call:

```sh
nvidia-smi --query-gpu=clocks.max.sm,memory.total,gpu_name --format=csv,noheader,nounits
```

This should give an output like this:

```
2100, 49140, NVIDIA RTX A6000
2100, 49140, NVIDIA RTX A6000
```

However, as seen on some [CERN GPU nodes](https://bigpanda.cern.ch/job?pandaid=7239692245), this is also possible:

```
[N/A], 24576, NVIDIA H100L-2-24C
```

Which is coming from a MIG partitioned GPU.

In this case we need to robustly handle this case, not failing, but printing a warning and setting GPU frequency to 0.

graeme-a-stewart:
GPU stats are obtained via an nvidia-smi call:

nvidia-smi --query-gpu=clocks.max.sm,memory.total,gpu_name --format=csv,noheader,nounits

That should give two numbers and a name.  However, as seen on some CERN GPU nodes, this is also possible:

```sh
[N/A], 24576, NVIDIA H100L-2-24C
```

Which is coming from a MIG partitioned GPU.

The parser will now handle this case, as well as other parsing failures, by printing a warning and setting the frequency to 0.

Closes #317.

### DIFF
--- a/package/src/nvidiamon.cpp
+++ b/package/src/nvidiamon.cpp
@@ -590,20 +590,16 @@ void const nvidiamon::get_hardware_info_smi(nlohmann::json& hw_json) {
   for (auto const& s : cmd_result.second) {
     std::istringstream instr(s);
     // Parse first field to a string in case it's weird, like "[N/A]"
+    // N.B. We expect this to be digits + a comma
     std::string tok;
     instr >> tok;
-    if (tok == "[N/A]") {
-      warning("Got [N/A] for GPU frequency");
+    errno = 0;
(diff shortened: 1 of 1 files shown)


amete (approved):
LGTM, thanks @graeme-a-stewart!

